Issue #4, opened 2019-03-06
As a developer, I need to provision shared Azure resources so that I can host and run an application.


rickrain:
## Description
Shared Azure resources are resources or services that are managed by a different group/team and have a different management life-cycle than application-specific resources or services.

This pattern assumes the following shared Azure resources and services:

- Azure Key Vault
- Virtual Network

## Requirements
- Azure CLI
- Terraform
- Key Vault
- Azure AD (AAD)

## Acceptance Criteria:
- A resource group is provisioned.
- A Key Vault resource is provisioned in the resource group.
- User can run a script from the command-line to deploy the template.

Given input company 'cblt'
and input location 'eastus'
when terraform apply is run
then a resource group in East US exists with the name 'core-infra-eus-cblt'
and a Key Vault exists with the name 'core-kv-eus-cblt' in resource group 'core-infra-eus-cblt'

## Notes
The resource group created in #30 can be repurposed for the resource group needed here.

code4clouds:
@rickrain  I think that instead of company we should use project to avoid collisions if two project groups within the same company implement this.

code4clouds:
@rickrain  I think that instead of company as an input we should use project to avoid collisions if two project groups within the same company implement this.

rickrain:
@code4clouds Agree.  We have refined this thinking and documented the naming convention standard in the wiki for **_future_** work against user stories.  However, for this user story please follow the AC as it is written and we'll add another enhancement issue for the next sprint.

PR #29, opened 2019-03-18
#4 Add script to deploy Resource group and Key Vault

manojvazirani:
#Steps 

Change into the ./cluster folder.

Run the script ./deploy.sh and pass the parameters if not set in environment variables as mentioned in README.md file.

This should deploy the resource group in azure and also the KeyVault in the same resource group.


xtophs:
what story (issue) is this PR implementing? 

Title says Key Vault. Terraform files provision ACR.

manojvazirani:
@xtophs : It is a work in progress [WIP] for the below Story to implement base infra using Terraform. PR was needed for pipeline checks to be validated.
https://github.com/Microsoft/entref-appservice-containers/issues/4

rickrain:
This WIP is not exactly consistent with the user story we're working against.  The pipeline will come in a later story as will permissions and policies around Key Vault.  The criteria is scoped to be able to run terraform apply to create the resource group and key vault - that's all.

manojvazirani:
@rickrain: The acceptance criteria for https://github.com/Microsoft/entref-appservice-containers/issues/4 mentions:
```An AAD service principal exists with permissions to read keys, certificates, and secrets from the Key Vault``` 
The policy and permissions were added based on the acceptance criteria above. We can modify the current PR based on the acceptance criteria or we can change the issue acceptance criteria based on our requirements for current story

**Comment by rickrain** [comment]
@manojvazirani, apologies for the confusion.  That part of the criteria should have been removed this morning when we were adjusting this user story.  A new user story for permissions and service principals will follow.

**Comment by xtophs** [comment]
A few comments before taking a closer look:

- Please separate out the repo init (.gitignore ) from the PR implementing #4.
- Azure DevOps pipeline was not an AC in #4. Please remove.
- Update `README.md` in the root to link to deployment instructions.

**Comment by code4clouds** [comment]
.gitignore is necessary to make sure no other users commit the .tfstates in the repo after running the command.

**Comment by code4clouds** [comment]
We do need a nice README.me story for the root level explaining the project @rickrain

**Comment by manojvazirani** [comment]
@xtophs @rickrain  The PR is up for review with updated documentation and integrated review comments.

**Comment by manojvazirani** [comment]
Rick,

Thanks for the above comments. 

- Slim this PR down to a single template that will deploy both the resources.
A single template is good to start with but as we are using modules, below link specifies how modules have to be structured, which is in sync with bedrock infrastructure. Do we want to make sure we have a folder structure defined, before making these changes?

- Remove the ./app folder. Terraform state files shouldn't be in the repo. These will be stored using remote backend storage in a later user story.
I was unable to find any ./app folder in my repo or changed files.

- I would argue that deploy.sh isn't needed (yet). Since it's prompting for things the user won't have, it can be replaced with terraform init and apply commands.
Agreed

- Rename ./cluster to `./shared'.
Agreed

**Comment by rickrain** [comment]
@manojvazirani Modules are used to group a set of resources together.  For example, the resources identified as shared resources (virtual network, key vault) could be packaged as a single module. The folder structure I have in mind is along these lines:

```
./shared
./app
```

This will become more evident and prescriptive in the upcoming user stories.  But the general idea is that you can go into the ./shared folder and deploy the virtual network and key vault.  Separately, you would be able to go into the ./app folder and deploy the application specific resources (app service plan, web app, etc.).

Apologies for the confusion around the `./app` folder.  That was on my system and I didn't properly cross check against your PR.  Thanks to @code4clouds for calling me out on this one.

**Comment by manojvazirani** [comment]
@rickrain : Below is the link based on which the current folder structure was defined: 
https://www.hashicorp.com/blog/hashicorp-terraform-modules-as-building-blocks-for
We can move to the above structure that you have in mind, once we define the overall folder structure which we plan on going forward, as discussed in the retro.

**Comment by charleszipp** [comment]
> @rickrain : Below is the link based on which the current folder structure was defined:
> https://www.hashicorp.com/blog/hashicorp-terraform-modules-as-building-blocks-for
> We can move to the above structure that you have in mind, once we define the overall folder structure which we plan on going forward, as discussed in the retro.

@manojvazirani What rick is suggesting is not counter to what is described in the article you provided. The module 'instance' thats used in that document is a logical concept not a GCP concept. That logical concepts abstracts what GCP resources implement that logical concept. 

In the past, I have done something similar for clusters. In past projects, I have worked with orgs to define what _compute_ means to them to build a generic _compute_ module. That may be made up of AKS or ASF or App Services or ACI. However, it wont be a single resource, there will be multiple resources that need to be provisioned (i.e. storage accounts, VM's or VMSS', etc). However, they all fit within the boundary of what the org considers its requirements for compute. 

What I am describing with compute example is what @rickrain is suggesting in this folder structure. What makes it confusing is we haven't found a good name for identifying the "shared" infrax that is used by basically every other module in the templates (think vnet, and key vault). 

I think what makes this even more confusing is bedrock, doesnt implement this logical module concept; which is the inspiration behind your initial implementation. Its just about demonstrating best practice for aks. Their logical modules are very high level AKS Single Region, Multi-Region, etc. Then their **logical modules** are composed of very granular **terraform modules** (i.e. they have a single module for key vault).

I would suggest you treat @rickrain's feedback as cannon to the guidance provided in the hashicorp article. The name shared is nebulous at best but, that's an easy change for the first person to suggest something better. **So forget the naming and semantics for now and treat it as we are advocating a logical abstraction over cloud-provider specific resources.**

**Comment by manojvazirani** [comment]
@charleszipp: I think you missed the original point for the conversation above, where @rickrain suggested combining the templates into one module. This goes against the structure that you are suggesting. Everything falls under shared(vnet, key vault, etc), I agree, but combining all resources in one template doesn't seem logical. This means that (resource group, vnet, keyvault) all fall inside the same template. The shared infrastructure was agreed on all all the time.

**Comment by charleszipp** [comment]
@manojvazirani @rickrain Here is the notes i was capturing as we were discussing that illustrates the expected structure, inputs, and resources naming.
```
./shared -> Infrax CI/CD Pipeline
	- variables.tf
		- app name input? -> defaults to core
		- location -> defaults to eastus
		- suffix -> dev-msft (no default)
	- main.tf
		- rg -> core-usea-rg-dev-msft
		- vnet -> core-usea-vnet-dev-msft
		- key vault -> core-usea-kv-dev-msft

./app -> App CI/CD Pipeline
	- variables.tf
		- app name -> cblt (no default)
		- location -> defaults to eastus
		- suffix -> dev-msft (no default)
	- main.tf
		- rg -> cblt-usea-rg-dev-msft
		- app services -> cblt-usea-app-dev-msft
		- web app -> cblt-usea-web-dev-msft
		- application gateway -> cblt-usea-agw-dev-msft
		- traffic manager -> cblt-usea-tfm-dev-msft
```

**Comment by code4clouds** [comment]
@charleszipp The infrastructure will be very complicated and it will be a very long file as @manojvazirani already learned while working on the ARM templates.  I think if we can do sub-modules, it will the right way to go and also recommended in "Terraform Up and Running" book.  http://shop.oreilly.com/product/0636920061939.do

**Comment by rickrain** [comment]
This is not a complicated infrastructure.  In ARM, it may be long because of the verbosity that JSON brings.  But, even in ARM it's a reasonable code base.  

Each of the folders (shared, app) could be packaged as modules.  So, we're on a path to do that if we want.  But, having a separate template for each individual resource is not a best practice for ARM or Terraform.

**Comment by charleszipp** [comment]
@code4clouds I am not disagreeing with the use of modules. However, I don't think we have enough clarification in our requirements to understand how best to segment our templates into modules. For example, if we are expected to share with bedrock, that could dramatically influence how we segment the templates. Additionally, we need more intel from walmart on the personas owning each part of the infrastructure as that could also influence how we split the modules.

I suggested to @manojvazirani that we take the simplest approach until we can get more clarification on the direction of this project (especially with its relationship to bedrock). In the spirit of not writing throw away code, I would suggest we avoid the overhead of plumbing the module variables and outputs until we have more clarification on requirements. cc @rickrain 

Hopefully that makes sense. We can discuss after standup if needed.

File: shared/README.md [hunk 1]
```diff
@@ -43,28 +42,9 @@ After saving the file set environment using:
...
-    description = "The short name of the application"
-    default = "cblt"
-}
-
-```
\ No newline at end of file
+Alternatively, use the cluster.tfvars file to set parameter values as shown below:
+
+``` 
+location="esatus"
```
xtophs: Please fix typo

manojvazirani: @xtophs: Sure will be part PR: https://github.com/Microsoft/entref-appservice-containers/pull/46 since it has to go through the entire review process to push into master.

Issue #30, opened 2019-03-18
As a developer or release engineer, I deploy terraform templates from my local machine, so that I can verify what the templates will provision.


charleszipp:
## Description
Intentionally lumping three things into this as they are all dependent on one another and each are relatively small in their effort. This will be focused on setting up the dev environment for the terraform templates. It will also establish the readme docs that explains how to execute the templates as a dev from vs code and as a release engineer from command prompt.

_TODO: inputs and resource group name may change. Rick trying to determine if there is a way to use a hash function for uniqueness in terraform. ARM supports this but terraform does not. This example uses an alternative method for uniqueness but results in longer names._

## Stories
- As an development engineer, I would like my terraform inputs to be different unique, so that my names do not overlap with other engineers
- As an development engineer, I do not want my terraform inputs to be checked into source control, so that other engineers do not modify my infrastructure
- As a release engineer, I would like to be able to deploy terraform templates using only a command line, so that I do not need to setup development tooling to deploy the templates

## Acceptance Criteria
- Development Engineers can define terraform inputs so that they do not have to be provided each time they run terraform validate, plan, or apply
- Development Engineers can define terraform inputs in isolation and those inputs are not checked into source control
- Release Engineers can deploy the terraform templates using only terraform and command line. Variables/inputs needed for this case should be documented within the readme.
- Resource group is deployed to demonstrate template does create resources in azure (see given/when/then scenario below)

given input app name 'cblt'
and input env 'dev'
and input location 'eastus'
and input org 'cse'
when terraform apply is run
then a resource group 'rg-cblt-dev-eus-cse' exists

PR #39, opened 2019-03-22
#30 Adding App Terraform Deployment

code4clouds:
Acceptance Criteria

    Development Engineers can define terraform inputs so that they do not have to be provided each time they run terraform validate, plan, or apply
    Development Engineers can define terraform inputs in isolation and those inputs are not checked into source control
    Release Engineers can deploy the terraform templates using only terraform and command line. Variables/inputs needed for this case should be documented within the readme.
    Resource group is deployed to demonstrate template does create resources in azure (see given/when/then scenario below)

given input app name 'cblt'
and input env 'dev'
and input location 'eastus'
and input org 'cse'
when terraform apply is run
then a resource group 'rg-cblt-dev-eus-cse' exists


code4clouds:
You need to use "sh deploy.sh"  deploy.sh by itself doesn't work without the +x   I didn't want the user to change permissions, but I found a way to commit my .sh with the +x so it should work on the next clone.  What was the error you got?

code4clouds:
The *.tfstate* may be only on your environment after you ran Terraform.  I haven't none checked in with this commit.

code4clouds:
@rickrain   "The intent of the user story is for this resource group to be used to define the core/shared resource (ie: Key Vault and Virtual Network) in later user stories."  There has to be some documentation for this code, right?

rickrain:
@code4clouds Yes, there will be documentation for this.  We will address this in an upcoming sprint.

code4clouds:
@rickrain Removed the ./README.md

We need to talk about the ./shared/README and discuss it further.

Issue #35, opened 2019-03-21
Use remote backend to track deployment state


charleszipp:
## Description
As an engineer, I would like to track what infrastructure is deployed to an environment from a shared/remote backend, so that the deployment changes can be determined from any machine.

### scenario 1: deployment with no existing resources
given a deployment is defined to deploy the following resources 

- resource group 'core-use-rg-dev-cblt'
- vnet 'core-use-rg-dev-cblt' in resource group 'core-use-rg-dev-cblt'
- key vault 'core-use-kv-dev-cblt' in resource group 'core-use-rg-dev-cblt'

and none of the above resources exist
when the deployment is executed
then the following resources are created

- resource group 'core-use-rg-dev-cblt'
- vnet 'core-use-rg-dev-cblt' in resource group 'core-use-rg-dev-cblt'
- key vault 'core-use-kv-dev-cblt' in resource group 'core-use-rg-dev-cblt'

### scenario 2: deployment with some existing resources
given a deployment is defined to deploy the following resources 

- resource group 'core-use-rg-dev-cblt'
- vnet 'core-use-rg-dev-cblt' in resource group 'core-use-rg-dev-cblt'
- key vault 'core-use-kv-dev-cblt' in resource group 'core-use-rg-dev-cblt'

and the following resources exist and were deployed from machine a

- resource group 'core-use-rg-dev-cblt'
- vnet 'core-use-rg-dev-cblt' in resource group 'core-use-rg-dev-cblt'

when the deployment is executed machine b
then the following resources are created

- key vault 'core-use-kv-dev-cblt' in resource group 'core-use-rg-dev-cblt'

### scenario 3: deployment with all resources existing
given a deployment is defined to deploy the following resources 

- resource group 'core-use-rg-dev-cblt'
- vnet 'core-use-rg-dev-cblt' in resource group 'core-use-rg-dev-cblt'
- key vault 'core-use-kv-dev-cblt' in resource group 'core-use-rg-dev-cblt'

and the following resources exist and were deployed from machine a
when the deployment is executed from machine b
then no new resources are created

## Notes
Use multiple directories (clone) to simulate executing the templates from different machines.

## Out of scope
Automating the creation of the storage account (remote backend)

PR #46, opened 2019-03-27
#35 - Add remote backend to track deployment state

manojvazirani:
PR for Issue: https://github.com/Microsoft/entref-appservice-containers/issues/35

As an engineer, I can now track what infrastructure is deployed to an environment from a shared/remote backend, so that the deployment changes can be determined from any machine.

This PR is dependent on the base test infrastructure in #45 as it adds unit tests to test backend infrastructure.


shared/main.tf
```diff
@@ -0,0 +1,73 @@
...
+    version = "~>1.21.0"
+}
+
+terraform {
+  required_version = "~> 0.11.11"
+  backend "azurerm" {
+  }
+}
+
+locals {
```
code4clouds: for readability can you indent this?

manojvazirani: Done automatically with merge from master.

day-jeff:
If you add me as a reviewer, I will be able to comment.

day-jeff:
> Please reference issue in description to have a link handy.
> 
> Please remember that this project is about teaching the consumer best practices. Neither Readme, nor PR description teach the cloning developers why this storage account is getting added.
> 
> @day-jeff for comments on how to best document.

Good catch, xtoph. Recommend that we re-use (as much as possible) text from this section of the Bedrock project: https://github.com/Microsoft/bedrock/tree/master/cluster/azure#configure-terraform-to-store-state-data-in-azure

TPark wrote the draft, and I edited it.

manojvazirani:
Thanks @xtophs and @day-jeff for the above review comments. Jeff, that did really help me with the storage account related documentation.

manojvazirani:
> Where is assert.sh? I don't see it in this PR or the current repo.

@xtophs As mentioned above, This PR is dependent on the base test infrastructure framework in #45 as it adds unit tests to test backend infrastructure. This PR is only going to be merged into master after PR #45. If there is a better way to do this, I am open to suggestions.

shared/README.md
```diff
@@ -7,25 +7,57 @@
 - Service Principal
```
rickrain: Service principal is not required here unless you're setting this up for CI which is not in scope.  Please remove.  When we get to adding CI, we will provide specific instructions that require this.

shared/README.md
```diff
@@ -7,25 +7,57 @@
...
+In order to setup an Azure backend, one needs an Azure Storage account.  If one must be provisioned, use [this](https://docs.microsoft.com/en-us/cli/azure/storage/account?view=azure-cli-latest#az-storage-account-create) link to provision, where `name` is the name of the storage account to store the Terraform state, `location` is the Azure region the storage account should be created in, and `resource group` is the name of the resource group to create the storage account in.  
+
+If there is already a pre-existing storage account, then retrieve the equivalent information for the existing account.
+
+Once the storage account details are known, we need to fetch the storage account key so we can configure Terraform with it:
+
+```bash
+>  az storage account keys list --account-name <storage account name>
+```
+
```
rickrain: The user also needs to have the storage account container created or identified at this step.  Please add instructions/comments for this.

shared/README.md
```diff
@@ -7,25 +7,57 @@
...
+With this, update `backend.tfvars` file in your cluster environment directory with these values and use `terraform init -backend-config=./backend.tfvars` to setup using the Azure backend.
```
rickrain: You need to `cd ./shared` before this step.  Also, suggest moving the `terraform init` command out of the paragraph and in a bash block like you do below.

shared/README.md
```diff
@@ -7,25 +7,57 @@
...
+
+### Execute the following commands to deploy resources:
 
 ``` bash
 $ cd ./shared
```
rickrain: Remove from this step since it's required before the previous step.

shared/README.md
```diff
@@ -42,9 +74,9 @@ After saving the file set environment using:
...
 ```
 
-Alternatively, use the cluster.tfvars file to set parameter values as shown below:
+Alternatively, use the shared.tfvars file to set parameter values as shown below:
 
 ``` 
-location="esatus"
+location="eastus"
 company="myCompany"
 ```
```
rickrain: Missing the following variable values:
- storage_account_name
- container_name
- access_key

shared/README.md
```diff
@@ -42,9 +74,9 @@ After saving the file set environment using:
 . .env
```
rickrain: Environment variables listed need to be consistent with variables shown in `.tfvars` file.  See comment below.

shared/backend.tfvars
```diff
@@ -0,0 +1,4 @@
+storage_account_name="storage-account-name"
+access_key="storage-account-access-key"
+container_name="storage-account-container"
+key="tfstate-key"
```
rickrain: Perhaps a better default name, such as `cobalt-test-terraform.tfstate` to indicate this is a tfstate file.  Something other than *-key in the name.

shared/README.md
manojvazirani: Its defined in the backend tfvars info. Those two are separate. Cannot be merged.

shared/README.md
manojvazirani: Its defined in the backend tfvars info. Those two are separate. Cannot be merged.

shared/README.md
```diff
@@ -4,28 +4,67 @@
...
 
-The following respources will be deployed
+The following resources will be deployed:
 - Azure Resource Group
 - Azure KeyVault 
 
 ## Deployment
 
-1. Authenticate using your Azure Principal or an Azure account with privileges to deploy resource groups.
+### Authenticate using your Azure Principal or an Azure account with privileges to deploy resource groups.
```
rickrain: Suggest changing "resource groups" to just "resources".

shared/README.md
```diff
@@ -4,28 +4,67 @@
...
 
 ``` bash
 $ az login
 ```
 
-2. Execute the following commands:
+### Configure Terraform to Store State Data in Azure
+
+Terraform records the information about what is created in a [Terraform state file](https://www.terraform.io/docs/state/) after it finishes applying.  By default, Terraform will create a file named `terraform.tfstate` in the directory where Terraform is applied.  Terraform needs this information so that it can be loaded when we need to know the state of the cluster for future modifications.
```
rickrain: Instead of _cluster_, let's use language like _infrastructure_.  _Cluster_ could be confusing to the reader in the context of this project.

shared/README.md
```diff
@@ -4,28 +4,67 @@
...
+
+In production scenarios, storing the state file on a local file system is not desired because typically you want to share the state between operators of the system.  Instead, we configure Terraform to store state remotely, and in Cobalt we use Azure Blob Store for this storage.  This is defined using a `backend` block.  The basic block looks like:
```
rickrain: Suggest changing _"...in Cobalt we use Azure Blob Store for this storage"_ to just _"in Cobalt we use Azure Blob Storage for this"_.

tests/backend_test/README.md
```diff
@@ -0,0 +1,36 @@
+# Backend Dev Testing
+
+## Description
+
+- main.sh is the parent file that contains all the tests
+- assert.sh is the code used for the base functions needed to test the backend.
```
rickrain: This file is one directory level up ( .. ).  Let's make it clear to the reader that this is a dependency for main.sh.  The way it reads now, the reader may be expecting this file to exist in the `backend_test` folder.

tests/backend_test/main.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -euo pipefail
+
+#. ../assert.sh continue_on_error #  load test functions with continue_on_error/stop_on_error
```
rickrain: Because this line is commented out, the tests fail to run.   Is the expectation that the user is running from the `backend_test` folder or the parent `test` folder?

tests/backend_test/main.sh
```diff
@@ -0,0 +1,49 @@
...
+assertResource core-usea-rg-test "Microsoft.KeyVault/vaults" core-usea-kv2-test not
+
+terraform init -backend-config=./test_backend.tfvars test2
+#Deploy Key Vault 2 only
+terraform apply -auto-approve test2
+
+assertRG core-usea-rg-test
+assertResource core-usea-rg-test "Microsoft.KeyVault/vaults" core-usea-kv1-test
+assertResource core-usea-rg-test "Microsoft.KeyVault/vaults" core-usea-kv2-test
+before=`az resource list -g core-usea-rg-test --resource-type "Microsoft.KeyVault/vaults"|jq .[].name| wc -l`
```
rickrain: The use of `jq `here is introducing a dependency that we don't need.  Please use the built-in `--query` feature of Azure CLI to query the results.  Reach out to me if you need any help constructing the query.

tests/backend_test/main.sh
```diff
@@ -0,0 +1,49 @@
...
+
+#Deploy Nothing since resources are already deployed
+terraform init -backend-config=./test_backend.tfvars test3
+terraform apply -auto-approve test3
+after=`az resource list -g core-usea-rg-test --resource-type "Microsoft.KeyVault/vaults"|jq .[].name| wc -l`
```
rickrain: Replace `jq` with Azure CLI's `--query`.

tests/backend_test/main.sh
```diff
@@ -0,0 +1,49 @@
...
+
+#. ../assert.sh continue_on_error #  load test functions with continue_on_error/stop_on_error
+
+#Cleanup
+terraform init -backend-config=./test_backend.tfvars test3
+terraform destroy -auto-approve test3
+
+#Check if resource group exists
+#Resource group does not exists
+assertRG core-usea-rg-test not
```
rickrain: We're essentially absorbing any assert failures with this approach.  For example, I create this resource group and then ran the test.  The output on the screen showed the test failed, but it then just continued successfully with the rest of the test.  In other words, $? reported 0 even though one of the tests failed.  If any test fails, the script should exit then with an exit code of 1.

tests/backend_test/README.md
manojvazirani: Yeah now there is only one main.sh to avoid confusion. Renamed the test file to backend_test.sh

tests/backend_test/main.sh
manojvazirani: Parent test folder. executed from main.sh

tests/backend_test/main.sh
manojvazirani: This is mainly because main.sh is currently executing as `assert.sh continue_on_error` and not `assert.sh stop_on_error`. Thats part of the readme under parent test folder from where the execution for backend tests is called from.

PR #52, opened 2019-04-09
Folder structure proposal to align cobalt closer with bedrock

erikschlegel:
**Description**
This is a proposal for the new cobalt folder structure. This structure was inspired from the [bedrock](https://github.com/Microsoft/bedrock/tree/master/cluster/azure) project and an attempt to standardize a cloud native turnkey infrastructure solution for application service scenarios which don't require a container orchestration cluster like kubernetes. 

This folder structure makes the following assumptions. 
- The cloud resource automation templates are written in the HashiCorp Configuration Language (HCL) for creating repeatable cloud resource declarations.
- The folder structure supports templates across multiple cloud providers. 
- The folder structure supports automated devops build and release pipelines across multiple providers such as Azure DevOPS, Jenkins, etc.
- There will be follow on documentation PR's for each folder's `REAMDE.md`. 

cobalt-architecture


code4clouds:
@erikschlegel Would you provide an example of what should be expected to be in the environments/internal-facing and environments/external-facing directories?

manojvazirani:
@erikschlegel : For resources like App gateway and API Manager which fall under different subnets, the subnet will be part of the same module?

erikschlegel:
@manojvazirani  No, they will be separate modules hence the folders `app-gateway` and `api-mgmt` 😄

manojvazirani:
> @manojvazirani No, they will be separate modules hence the folders `app-gateway` and `api-mgmt` 😄

@erikschlegel : Thanks, but my question was related to the subnets for each of the above resources. Looking at bedrock, the subnets are part of the resources.

erikschlegel:
> > @manojvazirani No, they will be separate modules hence the folders `app-gateway` and `api-mgmt` 😄
> 
> @erikschlegel : Thanks, but my question was related to the subnets for each of the above resources. Looking at bedrock, the subnets are part of the resources.

Both modules will reference the relevant subnet in the modules `main.tf` file. Similar to the `vnet` reference in bedrock [here](https://github.com/Microsoft/bedrock/blob/b1445116acdebbc5184976fa3782de4e40a0a91c/cluster/azure/vnet/main.tf#L15)

code4clouds:
@erikschlegel Bedrock is using environments to describe simple, cluster, and other types of deployments.  I see no internal vs external in their repo.

erikschlegel:
Those are examples of bedrock customer templates / environments.
Internal/external are templates specific to cobalt which have yet to be
defined. The architecture we reviewed yesterday is for multi-regional with
api management. We’ll use that as our starting template for cobalt and add
others once that template is running in production.

manojvazirani:
I see README's at a resource level than a module level. This will lead to multiple links where one Readme is pointing to another for information. My suggestion would be to reduce the number of readme files and aggregate the documentation at the parent directory level. Not particularly just one readme file but only at the parent levels.

erikschlegel:
> @erikschlegel Bedrock is using environments to describe simple, cluster, and other types of deployments. I see no internal vs external in their repo.

Those are examples of bedrock customer templates / environments.
Internal/external are templates specific to cobalt which have yet to be
defined. The architecture we reviewed yesterday is for multi-regional with
api management. We’ll use that as our starting template for cobalt and add
others once that template is running in production.

erikschlegel:
> I see README's at a resource level than a module level. This will lead to multiple links where one Readme is pointing to another for information. My suggestion would be to reduce the number of readme files and aggregate the documentation at the parent directory level. Not particularly just one readme file but only at the parent levels.

Readme's are provided down to the module level so we can provide detailed context around the intent, modules resources, usage, variables and output for each module.

Issue #58, opened 2019-04-12
As a developer, I have a Provider that configures Azure infrastructure for Terraform


James-Nance:
### Description
*As a developer, I have a Provider that configures Azure infrastructure for Terraform*
This should set up the basic Azure infrastructure to run Terraform.

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)
- [ ] Running on the correct Azure subscription
- [ ] Authenticates correctly
- [ ] …

### Resources
https://www.terraform.io/docs/providers/azurerm/index.html

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

PR #69, opened 2019-04-15
Add Provider that configures Azure infrastructure for Terraform

manojvazirani:
Added a provider that configures Azure infrastructure for Terraform. This sets up the basic Azure infrastructure to run Terraform. 

## All Submissions:

Issue Number: https://github.com/Microsoft/cobalt/issues/58

## What is the new behavior?

- This sets up the basic Azure infrastructure to run Terraform. 

- [NO]

## Other information

### DIFF
--- a/infra/modules/providers/azure/provider/README.md
+++ b/infra/modules/providers/azure/provider/README.md
@@ -10,14 +10,59 @@ This module will configure a provider and install required packages mentioned be
 
 Azure Resource Manager enables you to repeatedly deploy your app and have confidence your resources are deployed in a consistent state. In Cobalt, we use Azure Resource manager extensively for deployment of our resources using Terraform
 
+#### Usage
+
+Below code snippet example will use Azure Resource Manager to deploy a resource of type resource group. Additional details regarding Azure Resource Manager on Terraform can be found [here](https://www.terraform.io/docs/providers/azurerm/index.html)
+
+```
+resource "azurerm_resource_group" "example" {
+  name     = "${var.resource_group_name}"
+  location = "${var.resource_group_location}"
+}
+```
+
 ### Terraform (Version 0.11.13 or higher)
 
 Terraform enables you to safely and predictably create, change, and improve infrastructure. It is an open source tool that codifies APIs into declarative configuration files that can be shared amongst team members, treated as code, edited, reviewed, and versioned. In Cobalt, we use Terraform for all our resource deployments on Azure.
+
+#### Usage
+
+Below are the commands used for deploying resources on Azure using terraform templates. List of all terraform commands can be found [here](https://www.terraform.io/docs/commands/index.html)
+
+```
+terraform init  <folder-name>
+terraform apply <folder-name>
+```
 
 ### Azure Active Directory
 
 Azure Active Directory provides reliability and scalability one needs with identity services that work with on-premises, cloud, or hybrid environment. In Cobalt, Azure Active directory is used for the traffic manager role assignment.
 
+#### Usage
+
+Below code snippet example uses Azure Active Directory to Read service principal object to create a role assignment. Additional details regarding Azure Active Directory on Terraform can be found [here](https://www.terraform.io/docs/providers/azuread/index.html)
+
+```
+data "azuread_service_principal" "example" {
+    application_id = "${var.service_principal_id}"
+}
+```
+
 ### Null Provider 
 
 Null Provider provided by Terraform is needed in situations where one wants to execute external scripts to get configuration details of resources, not provided by terraform outputs, that are going to be created using terraform. In Cobalt, this is used mainly in test scripts, which do not actually have a resource defined but need a null resource to execute shell scripts.
+
+#### Usage
+
+Below code snippet example uses Null Provider to update trigger based on the trigger condition and executes the shell script locally. Additional details regarding Null Provider on Terraform can be found [here](https://www.terraform.io/docs/providers/null/index.html)
+
+```
+resource "null_resource" "example" {
+    triggers {
+        trigger = trigger-condition
+    }
+    provisioner "local-exec" {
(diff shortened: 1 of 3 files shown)


erikschlegel:
@manojvazirani how did you test this locally?

infra/modules/providers/azure/provider/README.md
```diff
@@ -0,0 +1,11 @@
+# Azure Provider
+
+## Summary
```
erikschlegel: This documentation seems really light. Can you add some more context? For example, which tricky behavior does the null provider assist with? Which limitations are you referring to?

erikschlegel: Here's an reference of the documentation posted for the lb terraform module https://github.com/Azure/terraform-azurerm-loadbalancer

manojvazirani: @erikschlegel : 

1. I know its light, but didn't want to mention the service principal stuff as its not part of the PR. The documentation mentions of all the details that are part of the PR. The service principal details should be part of the single-region PR.
2. Null provider, we are not really using now but its mainly to execute scripts as resources that are not defined. I dont think we are using it now so we can go either way, remove it for now and when null resource is needed, add `null provider` in the provider list at that time.
3. To test this locally, I did execute login using service principal, executed `terraform init` and then deployed a resource group using the above provider reference. I can add it part of the PR, but do we actually need that, as we will be creating actual resources and using this provider in the following stories.
```
az login --service-principal -u $APP_ID -p $APP_SECRET --tenant $TENANT_ID
```
```
module "provider" {
  source = "../provider"
}

resource "azurerm_resource_group" "cluster_rg" {
  name     = "cobalt-test-rg"
  location = "eastus"
}
```
I can add this file if you want it to be as part fo the current story.

infra/modules/providers/azure/provider/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Terraform enables you to safely and predictably create, change, and improve infrastructure. It is an open source tool that codifies APIs into declarative configuration files that can be shared amongst team members, treated as code, edited, reviewed, and versioned. In Cobalt, we use Terraform for all our resource deployments on Azure.
+
+### Azure Active Directory
+
+Azure Active Directory provides reliability and scalability one needs with identity services that work with on-premises, cloud, or hybrid environment. In Cobalt, Azure Active directory is used for the traffic manager role assignment.
+
+### Null Provider 
+
+Null Provider provided by Terraform is needed in various situations to help orchestrate tricky behavior or work around terraform limitations. In cobalt, this is used mainly in test scripts, which do not actually have a resource defined but need a null resource to execute shell scripts.
```
erikschlegel: can you replace `tricky behavior` with the technical challenge this module is intended to address?

Also, please provide more context to the `terraform limitations`

infra/modules/providers/azure/provider/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Azure Resource Manager enables you to repeatedly deploy your app and have confidence your resources are deployed in a consistent state. In Cobalt, we use Azure Resource manager extensively for deployment of our resources using Terraform
+
+### Terraform (Version 0.11.13 or higher)
+
+Terraform enables you to safely and predictably create, change, and improve infrastructure. It is an open source tool that codifies APIs into declarative configuration files that can be shared amongst team members, treated as code, edited, reviewed, and versioned. In Cobalt, we use Terraform for all our resource deployments on Azure.
+
+### Azure Active Directory
+
+Azure Active Directory provides reliability and scalability one needs with identity services that work with on-premises, cloud, or hybrid environment. In Cobalt, Azure Active directory is used for the traffic manager role assignment.
```
erikschlegel: Is this statement `Azure Active directory is used for the traffic manager role assignment.` accurate? Do we know for sure that the provider will only be used for traffic manager?

infra/modules/providers/azure/provider/README.md
```diff
@@ -0,0 +1,23 @@
...
+
```
erikschlegel: @manojvazirani please take a closer look at the example `REAMDE` I mentioned in my previous note. It would add more context for cobalt developers to include things like module usage, a bullet list of module goals, author, license, prerequisites, etc.

infra/modules/providers/azure/provider/README.md
manojvazirani: @erikschlegel: I believe we can remove Azure Active directory altogether based on the comments above since, currently in cobalt, we are not using AzureAD anywhere. But since bedrock was using it, I had to include it. 
When we reference modules from bedrock, will they be using the `bedrock-provider` or will it use the `cobalt-provider` as in that case both providers will be different from each other? This is what I mentioned in the call yesterday that needs some clarity. 
According to this example: https://github.com/Microsoft/bedrock/blob/master/cluster/environments/azure-simple
If Azure AD is not part of the `cobalt-provider`, deployments which depend on Azure AD are going to fail.

infra/modules/providers/azure/provider/README.md
manojvazirani: I agree on adding those for the actual modules which outputs as a resource but do we need to add those for provider module as they wont have any test scenarios. If we go that path, we should define a template for all the Readme's as a requirement for all the modules to be based on this:
https://github.com/Azure/terraform-azurerm-loadbalancer/blob/master/README.md

infra/modules/providers/azure/provider/main.tf
```diff
@@ -0,0 +1,15 @@
+provider "azurerm" {
+    version = "~>1.23.0"
+}
+
+terraform {
+  required_version = "~> 0.11.13"
```
erikschlegel: bedrock uses TF version `0.11.11` and so does our base image test harness.

Issue #61, opened 2019-04-12
As a developer, I can configure and deploy an Azure Service Plan


James-Nance:
### Description
*As a developer, I can configure and deploy an Azure Service Plan*

For Cobalt, Service Plan should define the resources to run a simple test app.

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)
- [ ] No hardcoded variables in module config; everything set through Terraform template
- [ ] Usage referenced in Azure Simple template
- [ ] Screenshot showing resource provisioned in Azure Portal

Also, here are a few points that need to be addressed:

1. No hardcoded variables in module config; everything set through Terraform template

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

PR #72, opened 2019-04-16
Configure and deploy an Azure Service Plan

manojvazirani:
As a user, I need to configure and deploy an Azure Service Plan. Without the App Service Plan, the app service cannot be deployed which is required in the following stories.

## All Submissions:

Issue Number: https://github.com/Microsoft/cobalt/issues/61

## What is the new behavior?

- This sets up the basic Azure infrastructure to run Terraform.
- Added the App Service Plan by using data inputs from the user in Terraform
- 

- [NO]

## Other information

This PR is dependent on the provider: PR https://github.com/Microsoft/cobalt/pull/69
cblt_svcplan


infra/modules/providers/azure/service-plan/variables.tf
```diff
@@ -0,0 +1,27 @@
+variable "resource_group_name" {
+  description = "Resource group name that the service plan will be created in."
+  default     = "cblt-svcplan-rg"
```
erikschlegel: the naming convention for a resource group should be more general ie remove `svcplan`

infra/modules/providers/azure/service-plan/variables.tf
```diff
@@ -0,0 +1,27 @@
...
+}
+
+variable "resource_group_location" {
+  description = "Default resource group location that the resource group will be created in. The full list of Azure regions can be found at https://azure.microsoft.com/regions"
+  type        = "string"
+}
+
+variable "svcplan_name" {
+    description = "The name of the servie plan to be created"
+    default   = "cblt-svcplan"
```
erikschlegel: if we're going to use `cblt` as a prefix then we should use the fullname.

infra/modules/providers/azure/service-plan/main.tf
```diff
@@ -0,0 +1,16 @@
...
+  name                = "${var.svcplan_name}"
+  location            = "${azurerm_resource_group.svcplan.location}"
+  resource_group_name = "${azurerm_resource_group.svcplan.name}"
+  kind                = "${var.svcplan_kind}"
+
+  sku {
+    tier = "${var.svcplan_tier}"
+    size = "${var.svcplan_size}"
+  }
+}
```
erikschlegel: This module should also support tag configuration... Here's an [example ](https://github.com/Microsoft/bedrock/blob/42047255990d34b95a359943304f7b5cea066fa7/cluster/azure/backend-state/resource-group.tf#L4)in bedrock.

infra/modules/providers/azure/service-plan/main.tf
```diff
@@ -0,0 +1,24 @@
...
+
+resource "azurerm_resource_group" "svcplan" {
+  name     = "${var.resource_group_name == "" ? "${local.name}-cobalt-rg" : "${var.resource_group_name}"}"
+  location = "${var.resource_group_location}"
+  tags     = "${merge(map("Name", "${local.name}"), var.resource_tags)}"
+}
+
+resource "azurerm_app_service_plan" "svcplan" {
```
erikschlegel: This module should also support other settings like `tags`, `capacity`, `reserved`. Keep in mind that `reserved` needs to be set to true for linux based app service plans. Here's a reference for all settings supported for the TF resource https://www.terraform.io/docs/providers/azurerm/r/app_service_plan.html#argument-reference

infra/modules/providers/azure/service-plan/README.md
```diff
@@ -0,0 +1,40 @@
+# Module Azure App Service Plan
+
+In App Service, an app runs in an App Service plan. An App Service plan defines a set of compute resources for a web app to run. These compute resources are analogous to the server farm in conventional web hosting. One or more apps can be configured to run on the same computing resources.
+
+This is a terraform module in Cobalt to provide an App Service Plan with the following characteristics:
+
+- Ability to specify resource group name in which the App Service Plan is deployed.
+- If a name is not specified, it will generate a random id and add it as a prefix for the names of all the resources created.
+- Ability to specify resource group location in which the App Service Plan is deployed.
+- Specify App Service Plan tier, size and kind of API manager to deploy
```
erikschlegel: Please provide the list of supported module settings including their description. Also include a reference to the terraform documentation https://www.terraform.io/docs/providers/azurerm/r/app_service_plan.html#argument-reference

infra/modules/providers/azure/service-plan/main.tf
```diff
@@ -0,0 +1,27 @@
...
+  name                = "${var.svcplan_name == "" ? "${local.name}-cobalt-svcplan" : "${var.svcplan_name}"}"
```
erikschlegel: The `name` should be driven solely on the `svcplan_name` var

infra/modules/providers/azure/service-plan/main.tf
```diff
@@ -0,0 +1,27 @@
+resource "random_id" "remotestate_account_name" {
+  byte_length = 6
+  keepers {
+    sa_account_ref = 1
+  }
+}
+
+resource "azurerm_resource_group" "svcplan" {
+  name     = "${var.resource_group_name == "" ? "${local.name}-cobalt-rg" : "${var.resource_group_name}"}"
```
erikschlegel: `name` should be driven directly off `resource_group_name `

infra/modules/providers/azure/service-plan/variables.tf
```diff
@@ -0,0 +1,47 @@
+variable "name" {
+  description = "(Required) Specifies the human consumable label for this resource."
```
erikschlegel: you may want to mention which resource names will be randomized in the absence of `name`

infra/modules/providers/azure/service-plan/main.tf
manojvazirani: It is a 3 step process and I agree on the way bedrock is doing it: 
1. Check if `svcplan_name` is provided, if provided assign it to name. 
2. Else check if `local.name` is provided by the user to be added as a prefix, 
3. if not provided then it will generate a randomized `local.name` prefix for the name.

infra/modules/providers/azure/service-plan/main.tf
manojvazirani: Check response as above.

infra/modules/providers/azure/service-plan/variables.tf
manojvazirani: This is provided in the output after the resources are created.

infra/modules/providers/azure/service-plan/main.tf
erikschlegel: This approach feels too opinionated for cobalt clients. This is a departure from the approach that bedrock [takes ](https://github.com/Microsoft/bedrock/blob/master/cluster/azure/aks/main.tf#L11)which contains no logic on aks name provisioning. I'd prefer to remove `local.name` all together and just `service_plan_name`.

infra/modules/providers/azure/service-plan/variables.tf
```diff
@@ -0,0 +1,47 @@
...
+  description = "(Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created. The full list of Azure regions can be found at https://azure.microsoft.com/regions"
+}
+
+variable "resource_tags" {
+  description = "Map of tags to apply to taggable resources in this module.  By default the taggable resources are tagged with the name defined above and this map is merged in"
+  type        = "map"
+  default     = {}
+}
+
+variable "svcplan_name" {
```
erikschlegel: As mentioned in previous comments, please avoid abbreviating resource names suggestion `service_plan_name`

infra/modules/providers/azure/service-plan/main.tf
```diff
@@ -0,0 +1,27 @@
+resource "random_id" "remotestate_account_name" {
```
erikschlegel: We should avoid business logic for randomizing resource names in terraform modules. This should be up to module clients. JFYI, [here's ](https://github.com/Microsoft/bedrock/blob/master/test/bedrock_Azure_common_kv_test.go#L18)how bedrock randomizes resource names for clusters.

infra/modules/providers/azure/service-plan/locals.tf
```diff
@@ -0,0 +1,3 @@
+locals {
```
erikschlegel: this file seems unnecessary

infra/modules/providers/azure/service-plan/variables.tf
manojvazirani: @erikschlegel: Seems like bedrock is not consistent with their approach of naming which is causing all this confusion and additional churn in project cobalt. I am looking at [this](https://github.com/Microsoft/bedrock/blob/42047255990d34b95a359943304f7b5cea066fa7/cluster/azure/backend-state/resource-group.tf) example which was provided by you [before](https://github.com/Microsoft/cobalt/pull/72#discussion_r276665806), as a reference to the changes needed, where they are using the exact same logic as us in cobalt for the naming of resources. If this feels opinionated for cobalt clients, we can go back to the approach that I had previously built without `local.name`.

erikschlegel: this comment was more related to avoid abbreviating resource names. Generally speaking, naming convention will vary from customer to customer. Let's start simple with supporting a single variable for service plan name and consider other edge cases based on feedback from our customers.

Issue #60, opened 2019-04-12
As a developer, I can configure and deploy Azure App Service


James-Nance:
### Description
*As a developer, I can configure and deploy Azure App Service*

For Cobalt, App Service should provide:

- App/container hosting
- Public IP mapped to container port (Dependency on Julio Research)
- Load balancing (Another User story for scale out config)

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)

- [ ] Service plan is already deployed before starting deployment of app service
- [ ] Key value pair for app service name and linux fx version
- open question: service aplan module depends on app service
- [ ]Output in readme with an example output
- [ ] VNET is an option for App Service. (No data block needed)

- [ ] No hardcoded variables in module config; everything set through Terraform template
- [ ] Usage referenced in Azure Simple template
- [ ] Screenshot showing resource provisioned in Azure Portal

Also, here are a few points that need to be addressed:

1. No hardcoded variables in module config; everything set through Terraform template

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

code4clouds:
App Service requires App Service Environment (ASE) to be in the same vnet as the AppGateway (If we are to follow the Game Plan).

code4clouds:
From the Game Plan diagram there is no public IP mapped on the App Service as the traffic is routed through the VNET as incoming from the App Gateway...    As in Traffic Manager-(public)->App Gateway-(vnet)->App Service.

code4clouds:
Should the key/secrets for ACR be wired to the Azure Key Vault?

James-Nance:
> Should the key/secrets for ACR be wired to the Azure Key Vault?

ACR and Key Vault would normally be managed by the dev / org, so not within scope of this (per the architecture diagram) infrastructure.  I'd say do what you need to do for testing purposes, but nothing should be hard-coded in the modules or templates.  Sorry if I'm misunderstanding the question.

[...]

PR #74, opened 2019-04-22
Configure and deploy an Azure App Service

manojvazirani:
As a developer, I can configure and deploy Azure App Service. The App service provides 
- App/container hosting using app service
- Public IP
- Load balancing

## All Submissions:

Issue Number: https://github.com/Microsoft/cobalt/issues/60

## What is the new behavior?

- Added the App Service Plan by using data inputs from the user in Terraform
- Added App Service by using data inputs from the user in Terraform
- Added Public IP and Load Balancer to the App Service module which uses the above inputs provided by the user.

- [NO]

## Other information

This PR extends on top of PR https://github.com/Microsoft/cobalt/pull/72

cobalt_appsvc


infra/modules/providers/azure/app-service/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+variable "name" {
+  default = "prod"
+}
+
+variable "location" {
+  default = "eastus"
+}
+
+resource "azurerm_app_service" "appsvc" {
```
erikschlegel: as mentioned in previous comments, the documented usage should be aligned with how we're using these modules in our own templates. That being said, we should reference our own modules as terraform modules.

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -0,0 +1,13 @@
+resource "azurerm_resource_group" "appsvc" {
```
erikschlegel: why are we defining a resource group here as opposed to referencing the one that contains the app service plan resource?

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -0,0 +1,13 @@
...
+  name     = "${var.resource_group_name}"
+  location = "${var.resource_group_location}"
+  tags     = "${var.resource_tags}"
+}
+
+resource "azurerm_app_service" "appsvc" {
+  name                = "${var.app_service_name}"
+  location            = "${azurerm_resource_group.appsvc.location}"
+  resource_group_name = "${azurerm_resource_group.appsvc.name}"
```
erikschlegel: to be clear, the resource group reference should be the one from the app service plan.

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -0,0 +1,13 @@
...
+  app_service_plan_id = "${var.app_service_plan_id}"
```
erikschlegel: the `app_settings` and `site_config` are required for container deployment... How would a developer provide a container and `DOCKER_REGISTRY_SERVER_URL` to the provisioned app service resource?

infra/modules/providers/azure/app-service/main.tf
manojvazirani: @erikschlegel: When we reference a resource group at resource level, and we have a case of multiple resources referencing the same resource, terraform deletes the resource and creates a new resource. But if its done at a module level, it works as expected. We did a couple different experimental runs at azure-simple level as bedrock is doing it and it works perfectly as expected. Thus we have to define the resource group at this level as opposed to referencing it.

infra/modules/providers/azure/app-service/main.tf
manojvazirani: The referencing happens at azure-simple level as compared to module definition level as explained above.

infra/modules/providers/azure/app-service/variables.tf
```diff
@@ -0,0 +1,46 @@
+variable "service_plan_resource_group_name" {
+  description = "The name of the resource group in which the service plan was created."
+  type        = "string"
+}
+
+
+variable "service_plan_name" {
+  description = "The name of the service plan"
```
erikschlegel: All variables should have a type

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -0,0 +1,28 @@
...
+resource "azurerm_app_service" "appsvc" {
+  name                = "${var.app_service_name[count.index]}"
+  resource_group_name = "${data.azurerm_resource_group.appsvc.name}"
+  location            = "${data.azurerm_resource_group.appsvc.location}"
+  app_service_plan_id = "${data.azurerm_app_service_plan.appsvc.id}"
+  tags                = "${var.resource_tags}"
+  count               = "${length(var.app_service_name)}"
+
+  app_settings {
+    DOCKER_REGISTRY_SERVER_URL      = "${var.docker_registry_server_url[count.index]}"
```
erikschlegel: This feels rigid as in most cases developers will be using the same docker registry. In the instance of deploying multiple app services, one would have to redefine the docker credentials each time.

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -0,0 +1,28 @@
...
+    DOCKER_REGISTRY_SERVER_USERNAME = "${var.docker_registry_server_username[count.index]}"
+    DOCKER_REGISTRY_SERVER_PASSWORD = "${var.docker_registry_server_password[count.index]}"
+  }
+
+  site_config {
+    linux_fx_version = "${var.site_config_linux_fx_version[count.index]}"
+    always_on        = "${var.site_config_always_on[count.index]}"
```
erikschlegel: This shouldn’t vary across multiple app services.

infra/modules/providers/azure/app-service/README.md
```diff
@@ -0,0 +1,60 @@
...
+  - DOCKER_REGISTRY_SERVER_URL : The docker registry server URL for app service to be created
+  - DOCKER_REGISTRY_SERVER_USERNAME : The docker registry server usernames for app services to be created
+  - DOCKER_REGISTRY_SERVER_PASSWORD : The docker registry server passwords for app services to be created
+- site_config : 
+  - linux_fx_version : Linux App Frameworks and versions for the App Services created. Possible options are a Docker container (DOCKER|<user/image:tag>), a base-64 encoded Docker Compose file (COMPOSE|${base64encode(file("compose.yml"))}) or a base-64 encoded Kubernetes Manifest (KUBE|${base64encode(file("kubernetes.yml"))}).
+  - always_on : Should the app be loaded at all times? Defaults to false.
+
+Please click the [link](https://www.terraform.io/docs/providers/azurerm/d/app_service.html) to get additional details on settings in Terraform for Azure App Service.
+
+## Usage
```
erikschlegel: Can you also include variable definitions or a snippet of your locals.tf.

infra/modules/providers/azure/app-service/README.md
```diff
@@ -0,0 +1,60 @@
...
+  count               = "${length(var.app_service_name)}"
+
+  app_settings {
+    DOCKER_REGISTRY_SERVER_URL      = "${var.docker_registry_server_url[count.index]}"
+    DOCKER_REGISTRY_SERVER_USERNAME = "${var.docker_registry_server_username[count.index]}"
+    DOCKER_REGISTRY_SERVER_PASSWORD = "${var.docker_registry_server_password[count.index]}"
+  }
+
+  site_config {
+    linux_fx_version = "${var.site_config_linux_fx_version[count.index]}"
```
erikschlegel: Where are these variables defined?

infra/modules/providers/azure/service-plan/variables.tf
```diff
@@ -19,18 +20,18 @@ variable "service_plan_name" {
 }
 
 variable "service_plan_tier" {
-    description = "The tier under which the service plan is created. Details can be found at https://docs.microsoft.com/en-us/azure/app-service/overview-hosting-plans"
-    default     = "Isolated"
+  description = "The tier under which the service plan is created. Details can be found at https://docs.microsoft.com/en-us/azure/app-service/overview-hosting-plans"
+  default     = "Isolated"
```
erikschlegel: This should not default to isolated as some developers will not need to run this on a private vnet. If a developer needs to run this in a vnet then we should provide a template to automate that and add this default in the relevant TF template(s).

infra/modules/providers/azure/service-plan/variables.tf
```diff
@@ -19,18 +20,18 @@ variable "service_plan_name" {
 }
 
 variable "service_plan_tier" {
-    description = "The tier under which the service plan is created. Details can be found at https://docs.microsoft.com/en-us/azure/app-service/overview-hosting-plans"
-    default     = "Isolated"
+  description = "The tier under which the service plan is created. Details can be found at https://docs.microsoft.com/en-us/azure/app-service/overview-hosting-plans"
```
erikschlegel: We have some variables with type and some without. Can we include them as a default across all our variable definitions.

infra/modules/providers/azure/app-service/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+  default     = ["", ""]
+}
+
+variable "docker_registry_server_password" {
+  description = "The docker registry server password for app service to be created"
+  default     = ["", ""]
+}
+
+variable "site_config_linux_fx_version" {
+  description = "The Linux App Framework and version for the App Service to be created"
```
erikschlegel: I feel like the  app service name and container should be merged as a single variable. Here’s an example for how this was done in a similar scenario https://link.medium.com/BDrofpzBcW

infra/modules/providers/azure/app-service/README.md
manojvazirani: The list above the Usage are the actual variables which are being used in the template.

infra/modules/providers/azure/app-service/README.md
manojvazirani: The module is being defined here. And the module usage is defined below.

infra/modules/providers/azure/app-service/README.md
manojvazirani: Variables are defined in variables.tf file

infra/modules/providers/azure/app-service/README.md
erikschlegel: where's `site_config_linux_fx_version` defined? There should be a clean separation between module usage and variable definition to avoid developer confusion.

infra/modules/providers/azure/app-service/main.tf
manojvazirani: It will be rigid, if we don't give developers the option. That is the reason each app-service has these settings.

infra/modules/providers/azure/app-service/README.md
erikschlegel: you should consider having a separate section in your readme with defining these variables.

infra/modules/providers/azure/app-service/main.tf
manojvazirani: That would be rigid since developers won't have the option. Anyway if its kept empty, all the values are going to be the same by default.

infra/modules/providers/azure/app-service/main.tf
erikschlegel: my point is that if you're using the same registry across 5 app services then someone is going to have to define the docker registry 5 times(all containing the same values)... Also, this seems like an unnecessary step for images hosted publicly on docker hub.

infra/modules/providers/azure/app-service/README.md
erikschlegel: both the app service name and container exec command should be driven off the same variable.

infra/modules/providers/azure/app-service/README.md
erikschlegel: the goal of these modules is to avoid developers from needing to define resource like this `resource "azurerm_app_service" "appsvc"`

infra/modules/providers/azure/app-service/variables.tf
manojvazirani: I believe this is an improvement task that needs to be added for future sprints.

infra/modules/providers/azure/service-plan/variables.tf
manojvazirani: This is what we had discussed in the meeting on call with @code4clouds. I did not have a default value before but we agreed to add this default for Service plans inside vnet based on the Game plan document.

infra/modules/providers/azure/app-service/README.md
manojvazirani: But wouldn't that mean having the same information in variables.tf and twice in the readme? 
Variables.tf already has a description of each variable defined in there.
While Readme has a description of each value needed in variables.tf.
What would be the use of `descriptions` in variables.tf if we have the same in descriptions at multiple places?

infra/modules/providers/azure/app-service/main.tf
manojvazirani: Yes but a developer now has the option of adding it for an app service. If not they are optional and can be kept empty. Presuming that the docker registry is going to be same is against best practices which kind of takes away the option of flexibility for the developer.

infra/modules/providers/azure/app-service/README.md
manojvazirani: But the actual usage is after the definition. A couple more lines down, the example usage specifies how it has to be used.

infra/modules/providers/azure/app-service/README.md
manojvazirani: In an ideal situation yes but that seems too rigid to push developers to give same names. Many containers will have additional versions that might not be part of the app service name.

infra/modules/providers/azure/app-service/main.tf
erikschlegel: not sure why this was marked as resolved as this approach would force me to provide my docker credentials for public images hosted on docker hub.

infra/modules/providers/azure/app-service/README.md
erikschlegel: ok.... if we go with multiple variables, how are you accounting for situations where there's a array length mismatch between `site_config_linux_fx_version` and `app_service_name` variables?

infra/modules/providers/azure/app-service/README.md
erikschlegel: my comment was not addressed. The documentation seems a bit confusing as this section represents how a developer would use the `app_service` module. Can you please remove the module definition from the readme and reference the reader to the location of the definition.

infra/modules/providers/azure/service-plan/variables.tf
erikschlegel: we discussed adding this default in the template that stands up a vnet, not the module. Most users of this module will not need to run this in isolated mode.

infra/modules/providers/azure/app-service/README.md
erikschlegel: Here's an usage documentation example for how the load balancer module is defined https://github.com/Azure/terraform-azurerm-loadbalancer

infra/modules/providers/azure/app-service/main.tf
erikschlegel: the variable name is`azurerm_resource_group.appsvc.name` which is misleading as the resource group name is sourced from the app service plan.

infra/modules/providers/azure/service-plan/variables.tf
code4clouds: This is a design decision that should be approved by @James-Nance @day-jeff .  If we are building an Enterprise level App Reference then Isolated should be mandatory as this is the most secure way to protect the information flowing between the services.  Enterprise App Reference in my head = PII, information that shouldn't be tampered or exposed outside of a VNET.

erikschlegel: Generally speaking, we still  need to do more research to double check whether isolated app service is required for private vnets / app services. This particular  module will be used by developers for app service plans which don’t necessarily require a private vnet. Isolated app services may make sense in certain scenarios but are  expensive [Instance](https://azure.microsoft.com/en-us/pricing/calculator/?service=app-service#app-service5dfb67a9-b49c-4c2c-82c4-9b726388f203). This decision is specific to the deployment environment(template) as this module will be used for  azure customer / developer(s) interested in deploying app services. In summary, private vnet is out of scope for this particular user story.

infra/modules/providers/azure/app-service/main.tf
manojvazirani: There are a couple of reasons for this approach:
1. The module app service will be consistent on the resource names.
2. While debugging an issue, module reference name will always help since every module will have a reference name.
3. If there are multiple resources referenced from previous modules, tracking back to where the resource came from will be easier for the user during deployment.

infra/modules/providers/azure/service-plan/variables.tf
```diff
@@ -19,18 +20,18 @@ variable "service_plan_name" {
...
-    description = "The compute and storage needed for the service plan to be deployed. Details can be found at https://azure.microsoft.com/en-us/pricing/details/app-service/windows/"
-    default     = "S1"
+  description = "The compute and storage needed for the service plan to be deployed. Details can be found at https://azure.microsoft.com/en-us/pricing/details/app-service/windows/"
+  default     = "S1"
 }
 
 variable "service_plan_kind" {
-    description = "The kind of Service Plan to be created. Possible values are Windows/Linux/FunctionApp/App"
-    default     = "Linux"
+  description = "The kind of Service Plan to be created. Possible values are Windows/Linux/FunctionApp/App"
```
erikschlegel: please include the variable type for all entries in this file so we're consistent with the other variable definitions.

infra/modules/providers/azure/app-service/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+variable "docker_registry_server_url" {
+  description = "The docker registry server URL for app service to be created"
+  type        = "list"
+  default     = ["", ""]
+}
+
+variable "docker_registry_server_username" {
+  description = "The docker registry server username for app service to be created"
+  type        = "list"
+  default     = ["", ""]
```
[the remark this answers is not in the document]
manojvazirani: Updated as discussed in the meeting

manojvazirani:
@erikschlegel @code4clouds : The PR is up for review again. 
- With respect to the open question that do we need a resource group name for service plan, the answer is yes since same service plan name can be in multiple resource groups. 
- So depends_on is not needed since service plan and resource group name is needed before deployment of app services which is now mentioned clearly in the documentation.

infra/modules/providers/azure/app-service/README.md
```diff
@@ -0,0 +1,69 @@
+## Azure App Service
+
+Build, deploy, and scale enterprise-grade web, mobile, and API apps running on any platform. Meet rigorous performance, scalability, security and compliance requirements while using a fully managed platform to perform infrastructure maintenance.
+
+More information for Azure App Services can be found [here](https://azure.microsoft.com/en-us/services/app-service/)
+
```
erikschlegel: General note, please update the usage section of the readme for the service plan module so that cobalt users have more context on how to define and configure there module.

manojvazirani: Sure, let me open a new PR with the documentation changes for service plan.

Issue #56, opened 2019-04-12
As a developer, I want to run integration tests for each CI build run


erikschlegel:
### Description
*As a developer, I want to run integration tests for each CI build run*
Running end to end integration tests to validate terraform deployments are operating as expected is essential for enterprises. Similar to bedrock, we'll use [terratest ](https://github.com/gruntwork-io/terratest)in the spirit of following azure devops [guidance](https://docs.microsoft.com/en-us/azure/terraform/terratest-in-terraform-modules) for terraform deployments. Initially, we'll run a light weight E2E test similar to bedrock's [azure-simple](https://github.com/Microsoft/bedrock/blob/master/test/bedrock_Azure_simple_test.go#L36) integration test. 

Creating the terraform template is out of scope for this feature. We'll use the app service terraform hello world sample [here](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/examples/app-service/linux-basic/main.tf) as our temporary test bed placeholder until the `azure-simple` template is released. This integration test should run terraform init, plan, format, destroy and apply on the mocked release and assert that it can establish a connection to the provisioned app service website with a 200 response.  

This azure devops build step will get triggered for pull request updates/submissions. The integration test harness should support both local and cloud runs. To simplify local dependency setup we'll use the docker base image from feature #55. We'll need to add a docker build step to the pipeline as the integration test(s) will run as a container.

### Acceptance Criteria
- [ ] Should assert that we have a mocked terraform template which our int test can apply which scaffolds a hello world app service website. 
- [ ] Should assert that we have a go terratest module that acts as an integration test which validates the response status code of the provisioned app service website.  
- [ ] Should assert that the go test run's within a docker container using `mage` and runs successfully locally and in azure devops. 
- [ ] Should assert that the `README` in the `test` directory is updated with local and cloud setup instructions.  

Also, here are a few points that need to be addressed:

1. Issues #55 and #54 are pre-req's

### Resources
*Mockups*
Include an animated GIF of the build running successfully in azDO

### Tasks
*Add the go test harness module to the `test` folder
*Add a mocked app service website terraform template within `environments`. This is a temporary solution to validate the test module until `azure-simple` is released. 
*Update the documentation in the test folder. 
*Dockerize the test harness run

PR #75, opened 2019-04-23
Erisch/features/devops/integration tests

erikschlegel:
## Test Automation Harness:
* [YES] Have you added an explanation of what your changes do and why you'd like us to include them

**Description**
This pull request adds a new automated test harness to account for unit and integration tests. We recommend cobalt contributors provide automated test artifacts in the `test` folder  for each terraform template. We're using `mage` to orchestrate the test run. This harness supports integration test classes where thew class name starts with `TestIT` and unit tests that start with `TestUT`.

The test runner will automatically run integration and unit tests based on the git change log of the current working branch. We took a naive approach with re-running all template tests if TF module artifacts are modified. We'll follow up a with a PR to build a module reference <> template dependency graph to follow a more robust and sophisticated approach. 

We also put a focus on developer experience with simplifying local automated test runs with including a dockerized test runner. The dockerized test harness will be migrated to azure devops once the base image is multi staged. 

**Feature List**
- Integrating the test harness with our Azure DevOPS CI pipeline. 
- A dockerized test harness that can be invoked via `./test-harness/local-run.sh`
- Excluding unmodified terraform templates from test run
- An environment template file `.env.template`
- `dep` manifest for terratest dependencies. 
- Base docker image
- An nginx container hello world app service terraform template `azure-simple-hw`
- A backend template for setting up remote terraform state resources in azure. This template references the `backend-state` module from bedrock. 
- An golang based integration test for the hello world template. This test asserts that an http connection can be established to the app service resource.
- Removed unused files. 
- A shell script utility to automate the base image build step. 
- The `magefile.go` test orchestrator.
- Added a lint format step to our test runner. 
- Added key vault integration with our azure devops pipeline.

[NO] Does this include new/updated documentation?
**Details**
 To avoid tacking more changes onto this mounting PR I opted to submit a follow on PR that focuses on documentation. The user story will be closed once both PR's are merged. 

[YES] I have added tests to cover my changes.
[YES] All new and existing tests passed.
[YES] My code follows the code style of this project.
[YES] I ran lint checks locally prior to submission.
[YES] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Issue Number: #56 

- NO

[Logs](https://dev.azure.com/csedallascrew/project-cobalt/_build/results?buildId=159) from passing CI build from azure DevOPS

image


infra/templates/backend-state-setup/terraform.tfvars
```diff
@@ -0,0 +1,2 @@
+name="cobaltfstate"
+location="<az-resource-group-location>"
```
nmiodice: How does this value get inflated? Seems easy to miss and accidentally get this default string as our location.

infra/templates/backend-state-setup/variables.tf
```diff
@@ -0,0 +1,17 @@
+variable "name" {
+  type        = "string"
+  description = "(Required) Specifies the human consumable label for this resource."
+  default     = ""
```
nmiodice: You may want to instead omit the default completely. This will cause the TF commands to fail unless its provided as an argument explicitly. 

similar comment for other variables in the project.

infra/templates/backend-state-setup/variables.tf
```diff
@@ -0,0 +1,17 @@
...
+}
+
+variable "location" {
+  type        = "string"
+  description = " (Required) Specifies the supported Azure location where the resource exists. Changing this forces a new resource to be created."
```
nmiodice: required is redundant as the TF will fail with a similar note if its not provided as an arg

magefile.go
```diff
@@ -0,0 +1,71 @@
...
+
+// The default target when the command executes `mage` in Cloud Shell
+var Default = Full
+
+func main() {
+	Default()
+}
+
+// A build step that runs Clean, Format, Unit and Integration in sequence
+func Full() {
```
nmiodice: nit: more clear if named something like `FullTestSuite` or `RunAllTests`

Similar -- `Unit` and `Integration` may be more clear if its `RunUnitTests` or `RunIntegrationTests`

... total nit. ignore if you want.

test-harness/build-base-image.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Arguments
+declare go_version=""
+declare tf_version=""
+declare docker_img="msftcse/cobalt-test-base"
+declare docker_file="test-harness\docker\base-images\Dockerfile"
```
nmiodice: Why are we using `/` and `\`?

test-harness/Dockerfile
```diff
@@ -0,0 +1,15 @@
+ARG base_img_tag
+
```
nmiodice: Can you add comments about the usage of this file vs `test-harness/docker/base-images/Dockerfile`?

test-harness/init.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+declare -A TEST_RUN_MAP
+declare readonly TEMPLATE_DIR="infra/templates"
+
+function check_required_env_variables() {
```
nmiodice: could be shortened using a for loop that consumes a list of args

test-harness/init.sh
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
nmiodice: Can you add a documentation here that explains the use of this file? I see what its doing, but its complicated enough that some guidance would help IMO

test-harness/local-run.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
```
nmiodice: Thank you for making local testing easier!!

Unsure how close we can get here, but what do you see as the absolute closest we can have our local testing match the testing from the pipeline steps? This is always a huge pain point that I have. If you could solve it (perhaps this approach already does) it would be a huge win IMO.

infra/templates/azure-simple-hw/test/azure_simple_integration_test.go
```diff
@@ -0,0 +1,100 @@
+package test
```
nmiodice: This looks like a fair bit of code that is unrelated to the assertions you want to make. Have you considered introducing a layer of abstraction that could be used to make the tests standardized and easy to write?

Gopkg.toml
```diff
@@ -0,0 +1,42 @@
+# Gopkg.toml example
```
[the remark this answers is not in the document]
erikschlegel: I'm still a golang noob and not sure how much control we have on these files as they were auto generated.

devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -46,3 +68,37 @@ steps:
...
+- script: |
+    go run $TEST_HARNESS_GOFILE
+  env:
+    ARM_CLIENT_ID: $(AD-SP-CLIENT-ID)
+    ARM_CLIENT_SECRET: $(AD-SP-SECRET)
+    ARM_SUBSCRIPTION_ID: $(AD-SP-SUBSCRIPTION-ID)
+    ARM_TENANT_ID: $(AD-SP-TENANT-ID)
+    ARM_ACCESS_KEY: $(ARM-ACCESS-KEY)
+  workingDirectory: '$(modulePath)/$(BUILD_TEMPLATE_DIRS)'
+  displayName: 'Run Unit + Integration tests'
```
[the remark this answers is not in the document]
erikschlegel: This is a valid point. We could split things out where we run unit tests in the CI and integration tests in the deployment release. Running both unit and integration in the same CI matches the bedrock flow. Whichever we decide, we should be consistent across both projects. @NathanielRose thoughts?

go.mod
```diff
@@ -0,0 +1,68 @@
+module cobalt
+
+require (
```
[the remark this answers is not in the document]
erikschlegel: great question. This is auto pulled in through the terratest `imports`

```
	http_helper "github.com/gruntwork-io/terratest/modules/http-helper"
	random "github.com/gruntwork-io/terratest/modules/random"
	terraform "github.com/gruntwork-io/terratest/modules/terraform"
	test_structure "github.com/gruntwork-io/terratest/modules/test-structure"
	terraformCore "github.com/hashicorp/terraform/terraform"
```

infra/templates/azure-simple-hw/test/azure_simple_integration_test.go
erikschlegel: that's a good point. As per our side chat this will be a top focus for our next sprint.

test-harness/local-run.sh
erikschlegel: at a minimum, both `local-run` and `azure-pipelines` should use the same test harness container(that's coming soon with multi stage builds). I'd have to think about an effective way to locally replicate the branch naming convention azure devops follows.

.env.template
```diff
@@ -0,0 +1,16 @@
+BUILD_BUILDID=1
+BUILD_UPSTREAMBRANCH=master
+BUILD_SOURCEBRANCHNAME=`git branch | sed -n '/\* /s///p'`
```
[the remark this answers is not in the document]
erikschlegel: that's a great point. I'll move those to `local-run.sh`.

devops/providers/azure-devops/templates/azure-pipelines.yml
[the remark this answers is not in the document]
nmiodice: As per the discussion we had today I'm OK leaving this as is. It seems like there will be future work to extend and standardized the testing patterns between the project as more requirements are gathered.

magefile.go
```diff
@@ -0,0 +1,71 @@
...
+// A build step that runs integration tests
+func RunIntegrationTests() error {
+	mg.Deps(CleanAll)
+	mg.Deps(FormatGolangFiles)
+	fmt.Println("INFO: Running integration tests...")
+	return sh.RunV("go", "test", "./...", "-run", "TestIT", "-v", "-timeout", "99999s")
+}
+
+// A build step that formats both Terraform code and Go code
+func FormatGolangFiles() error {
```
nmiodice: [question for the future] Should this be caught by linting and cause a build to fail? I think you may be able to run something like this to check if the terraform is correctly formatted:

```
# https://www.terraform.io/docs/commands/fmt.html
terraform fmt -check=true
```

infra/templates/azure-simple-hw/test/azure_simple_integration_test.go
```diff
@@ -0,0 +1,113 @@
...
+		BackendConfig: map[string]interface{}{
+			"storage_account_name": tfstate_storage_account,
+			"container_name":       tf_state_container,
+		},
+	}
+
+	return terraformOptions
+}
+
+const SKIP_STAGE_ENV_VAR_PREFIX = "SKIP_"
```
nmiodice: What is the use case for this feature?

infra/templates/backend-state-setup/terraform.tfvars
```diff
@@ -0,0 +1,2 @@
+name="cobaltfstate"
```
nmiodice: As per our discussion, are we planning on preferring `variables.tf` or CLI args to `*.tfvars`?

magefile.go
erikschlegel: Agree that we should create a user story to support a proper TF Lint target within our test harness.

devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -20,29 +22,64 @@ pr:
...
 
 steps:
+- checkout: self
+  persistCredentials: true
+  clean: true
+
+- script: |
+    docker login $ACR_HOST -u $ACR_USERNAME -p $(DOCKER-PASSWORD)
+
+- script: |
```
NathanielRose: So checked with AzDO there isnt a way to trigger tasks from repo path and this is a good approach to acheive this. Another approach would have been to have multiple build jobs based on repo file path  that would execute different parts of the test suite. Only concern here is it may get tricky to get the logs out running everything inside docker.

PR #78, opened 2019-04-26
Integration test fix for branch merges

erikschlegel:
## All Submissions:

## Hot Fix Description
This hot fix addresses an issue where feature merge branches to master cause the build to fail because we're comparing master to master. The new behavior is that we compare `master` <> `master~` to determine the `git diff` for merges into master. 

This also happened to surface other issues which this PR addresses. The other issue this addresses is that changes within the `test-harness` directory were being ignored for to default the test harness to run all templates.

I also moved the integration test to a new folder called `integration` so we have a clean separation between unit and integration tests. I also added a new terraform variable `app_service_linux_container_command` to manage the linux container exec command on the app service. 

Issue Number: #56 

- [NO]


devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -61,10 +61,11 @@ steps:
     BASE_IMAGE=$ACR_HOST/$TEST_HARNESS_BASE_IMAGE:$TEST_HARNESS_BASE_IMAGE_TAG
     substitution="remotes/origin"
     replace_string="refs/heads"
+    GIT_DIFF_UPSTREAMBRANCH="remotes/origin/master"
     GIT_DIFF_SOURCEBRANCH=`[[ -n $SYSTEM_PULLREQUEST_SOURCEBRANCH ]] && \
         echo "${substitution}/${SYSTEM_PULLREQUEST_SOURCEBRANCH}" || \
         echo "${BUILD_SOURCEBRANCH//$replace_string/$substitution}"`
-    GIT_DIFF_UPSTREAMBRANCH="remotes/origin/master"
+    [ $GIT_DIFF_SOURCEBRANCH == $GIT_DIFF_UPSTREAMBRANCH ] && { GIT_DIFF_SOURCEBRANCH="$GIT_DIFF_SOURCEBRANCH~"; }
```
nmiodice: Reading up on this, I suspect that this strategy will only work for squash merge PR. This is because in a squash merge PR, the whole commit history is converted into a single commit, so you can be sure that `GIT_DIFF_SOURCEBRANCH` is indeed `master~`.

If a non-squash merge is used, then each individual commit that the requester made will be added, so you don't know how many commits to look back when determining `GIT_DIFF_SOURCEBRANCH`. 

More info here (note, the images are extremely unclear here... read the text as it is more relevant.):
https://help.github.com/en/articles/about-pull-request-merges

Are we enforcing by some programmatic method that PRs are squash merge PRs?

devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -73,13 +74,15 @@ steps:
...
-      docker run -e ARM_SUBSCRIPTION_ID=$(AD-SP-SUBSCRIPTION-ID) \
-          -e ARM_CLIENT_ID=$(AD-SP-CLIENT-ID) \
-          -e ARM_CLIENT_SECRET=$(AD-SP-SECRET) \
-          -e ARM_TENANT_ID=$(AD-SP-TENANT-ID) \
-          -e DATACENTER_LOCATION=$DATACENTER_LOCATION \
-          -e TF_STATE_STORAGE_ACCT=$TF_STATE_STORAGE_ACCT \
-          -e TF_STATE_CONTAINER=$TF_STATE_CONTAINER \
-          -e ARM_ACCESS_KEY=$(ARM-ACCESS-KEY) \
-          --rm $BUILD_TEST_RUN_IMAGE:$BUILD_BUILDID
+    if test ! -z "$(docker images -q ${BUILD_TEST_RUN_IMAGE}:${BUILD_BUILDID})"; then
```
nmiodice: When would the docker image not exist on the host? Do we want to fail in that case, or is there a use case I'm not aware of?

Also, TIL about `test`. Thanks!

test-harness/init.sh
```diff
@@ -47,7 +47,8 @@ function default_to_all_template_paths() {
 
 function add_template_if_not_exists() {
     declare readonly template_name=$1
-    if [[ -z ${TEST_RUN_MAP[$template_name]+unset} ]]; then
+    declare readonly template_directory="$TEMPLATE_DIR/$template_name"
+    if [[ -z ${TEST_RUN_MAP[$template_name]+unset} && -d "$template_directory" ]]; then
         TEST_RUN_MAP[$template_name]="$TEMPLATE_DIR/$template_name"
```
nmiodice: replace `"$TEMPLATE_DIR/$template_name"` with the already created `"$template_directory"`

test-harness/init.sh
```diff
@@ -77,17 +78,21 @@ function template_build_targets() {
     [[ -z $GIT_DIFF_SOURCEBRANCH ]] && echo "ERROR: GIT_DIFF_SOURCEBRANCH wasn't provided" && return 1
 
     echo "INFO: Running git diff from branch ${GIT_DIFF_SOURCEBRANCH}"
-    files=(`git diff ${GIT_DIFF_UPSTREAMBRANCH} ${GIT_DIFF_SOURCEBRANCH} --name-only|grep -v *.md`)
+    files=(`git diff ${GIT_DIFF_UPSTREAMBRANCH} ${GIT_DIFF_SOURCEBRANCH} --name-only|grep -v *.md||true`)
```
nmiodice: should we instead prefer a white-list of file extensions? The list would probably include `.tf` and other terraform related files. May be more robust this way so that the tests won't be re-run if somebody adds, for example, an image or text file.

test-harness/init.sh
```diff
@@ -97,5 +102,10 @@ function template_build_targets() {
         esac
     done
 
+    if [ ${#TEST_RUN_MAP[@]} -eq 0 ]; then
+        echo "INFO: No templates to process. Exiting build step"
+        exit 0
+    fi
```
nmiodice: Is this necessary? Without it, will the tests harness just loop over an empty array and produce no results anyways?

test-harness/init.sh
```diff
@@ -77,17 +78,21 @@ function template_build_targets() {
...
+    files=(`git diff ${GIT_DIFF_UPSTREAMBRANCH} ${GIT_DIFF_SOURCEBRANCH} --name-only|grep -v *.md||true`)
     for file in "${files[@]}"
     do
         IFS='/' read -a folder_array <<< "${file}"
-        
         if [ ${#folder_array[@]} -lt 3 ]; then
             continue
         fi
 
+        if [ ${folder_array[0]}=='test-harness' ]; then
```
nmiodice: Can you explain the special index values of `0` and `1`? How do we know that these are the right indices to look for these folder names?

erikschlegel: sure... So if you think about the folder parts of the `PWD`. If any file(s) happen to change within `/test-harness` or `/infra/modules` then the harness runs all tests within the `/infra/templates` should be carried over to the build directory for the run via `default_to_all_template_paths`. It's a naive approach and we have a user story to design a more elegant algo based off a module dependency graph.

test-harness/init.sh
erikschlegel: if there's no template directories to include in the run then we should exit and skip rebuilding the docker image.

devops/providers/azure-devops/templates/azure-pipelines.yml
erikschlegel: if there's no template directories to process then we wouldn't have built the image.

devops/providers/azure-devops/templates/azure-pipelines.yml
erikschlegel: cobalt has enabled [this ](https://help.github.com/en/articles/configuring-commit-squashing-for-pull-requests) which forces developers to squash merges

test-harness/init.sh
erikschlegel: I'm going to add the following 
```
GIT_DIFF_EXTENSION_FILTER="*.md"
files=(`git diff ${GIT_DIFF_UPSTREAMBRANCH} ${GIT_DIFF_SOURCEBRANCH} --name-only|grep -v ${GIT_DIFF_EXTENSION_FILTER}||true`)
```

if you want to support multiple extensions then you can do something like this 
GIT_DIFF_EXTENSION_FILTER="*.md|*.txt"

nmiodice: Seems like we care only about when Terraform files change, so why not explicitly filter for those (i.e., `grep` and not `grep -v`? With the current logic, adding a `.txt` file to each directory would cause a massive integration test to run.

I think a white-list approach is less brittle. Though what you have won't break anything, so I suppose its not a blocking comment.

test-harness/init.sh
nmiodice: Makes sense, thanks for clarifying

devops/providers/azure-devops/templates/azure-pipelines.yml
nmiodice: Awesome!

test-harness/init.sh
erikschlegel: That’s a good call out. We’d have to account for changes to docker files or the test harness shell script(s) and go filed... I’m thinking the pattern would look something like this “*.go|*.tf|*.sh|Dockerfile*|*.tfvars”.

Issue #109, opened 2019-05-17
[Testing] Implement integration test capabilities for Test Framework component


nmiodice:
### Description
*As a template maintainer, I'd like to have a framework that I can use to simplify writing integration tests in order to reduce the friction to writing robust tests for my templates.*

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)
- [x] Test framework should be augmented to support integration tests. The patterns to follow are shown in [Story #90](https://github.com/microsoft/cobalt/issues/90)
- [x] `azure-simple-hw` integration test is updated to use the new harness
- [x] `README.md` in test harness is updated to include references and examples of the integ test functionality.
- [x] Demo showing how to write an integ test is produced

### Resources
N/A

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

nmiodice:
From #90 
Pattern for integration test implementation

```go
// Something like this can be used to hold a reference to:
//    - Go Test
//    - Terraform Options
//    - List of user-defined functions that can be run to validate the deployment
//    - etc...
type IntegrationTestFixture struct {
	GoTest                *testing.T         // Go test harness
	TfOptions             *terraform.Options // Terraform options
        ...
}

// Something like this can be used to manage the terraform and test lifecycle
func RunIntegrationTests(fixture *IntegrationTestFixture ) {
}
```

Clients can use this using code similar to:
```go
func TestFoo(t *testing.T) {
	test_fixture := infratests.IntegrationTestFixture{
		GoTest:                t,
		TfOptions:             buildTerraformOptions(),
                ...
	}

	infratests.RunIntegrationTests(&test_fixture)
}
```

nmiodice:
PR: https://github.com/microsoft/cobalt/pull/129

PR #129, opened 2019-05-22
Add integ. test support to test harness

nmiodice:
## All Submissions:

Integration tests are cumbersome to write. The way that we write them is not standardized.

Issue Number: https://github.com/microsoft/cobalt/issues/109

## What is the new behavior?
Implement integ test support to the testing framework.

- YES

Users of the test harness will need to re-build the test image and/or pull down the latest image.


test-harness/infratests/unit.go
```diff
@@ -20,7 +17,8 @@ import (
 	terraformCore "github.com/hashicorp/terraform/terraform"
 )
 
-type ResourceAttributeValueMapping map[string]map[string]string
+type AttributeValueMapping map[string]string
```
erikschlegel: remind me again why we're only supporting string comparison for our plan reconciliation check?

test-harness/Dockerfile
```diff
@@ -6,5 +6,9 @@ ARG build_directory
 RUN echo "INFO: copying $build_directory"
 # Copy the recently modified terraform templates
 ADD $build_directory *.go ./
+
+# TODO: when this is moved into the orion repo the following line can be deleted
+ADD test-harness/ ./test-harness
```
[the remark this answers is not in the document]
nmiodice: This is the whole test harness + the magefile. It copies over the test package (which will later be included via go mod) as well.

Once we can use go mod to install the dependency we can rethink the approach

infra/templates/azure-simple-hw/tests/integration/azure_simple_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"testing"
+	"time"
+
+	httpClient "github.com/gruntwork-io/terratest/modules/http-helper"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/microsoft/cobalt/test-harness/infratests"
+)
+
+var prefix = fmt.Sprintf("cobalt-%s", random.UniqueId())
```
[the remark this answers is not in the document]
nmiodice: Can investigate though I just want to call out that the decision here is something that will be done on a test by test basis. Don't see how to internalize to the test harness, do you see a good way?

test-harness/infratests/integration.go
```diff
@@ -0,0 +1,117 @@
...
+//	- Create new terraform workspace. This helps prevent accidentally deleting resources
+//	- Run `terraform apply`
+//	- Run `terraform output`
+//	- Validate outputs
+//	- Run user-supplied validation of outputs
+//	- Destroy resources
+func RunIntegrationTests(fixture *IntegrationTestFixture) {
+	terraform.Init(fixture.GoTest, fixture.TfOptions)
+
+	workspaceName := random.UniqueId()
```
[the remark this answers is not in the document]
nmiodice: I could see issues where a Dev uses workspace FOO for testing and accidentally would blow away their workspace by running an integration test. 

The benefit of this approach is that each run is isolated fully.

Thoughts?

test-harness/infratests/integration.go
```diff
@@ -0,0 +1,117 @@
...
+//	  output has more mappings
+//	- The output has the correct number of items
+//	- Run any user-supplied assertions over the output
+func validateTerraformOutput(fixture *IntegrationTestFixture, output TerraformOutput) {
+	validateTerraformOutputCount(fixture, output)
+	validateTerraformOutputKeyValues(fixture, output)
+
+	// run user-provided assertions over the TF output
+	for _, outputAssertion := range fixture.TfOutputAssertions {
+		outputAssertion(fixture.GoTest, output)
```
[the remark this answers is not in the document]
nmiodice: The convention is the same as go testing in general - if the test finds something abnormal it calls test.Fatal(some error). The user defined function is responsible for this.

However the test I have as an example does not do that so I need to fix it! Thanks for the comment here!

test-harness/infratests/unit.go
nmiodice: Everything in terraform 11 is flattened to a string in the plan. List indexes become part if the attribute key, as an example

test-harness/Dockerfile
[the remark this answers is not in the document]
nmiodice: A slightly more effecient way is to just copy the magefile and go package but the benefit is a few KB max

infra/templates/azure-simple-hw/tests/integration/azure_simple_test.go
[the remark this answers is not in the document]
nmiodice: discussed offline to use a more descriptive prefix that does not use the workspace

test-harness/infratests/integration.go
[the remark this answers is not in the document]
nmiodice: discussed offline that random workspace is OK but that the prefix will be more descriptive

test-harness/infratests/unit.go
```diff
@@ -42,19 +40,19 @@ type UnitTestFixture struct {
 func RunUnitTests(fixture *UnitTestFixture) {
 	terraform.Init(fixture.GoTest, fixture.TfOptions)
 
-	workspace_name := random.UniqueId()
-	terraform.WorkspaceSelectOrNew(fixture.GoTest, fixture.TfOptions, workspace_name)
-	defer terraform.RunTerraformCommand(fixture.GoTest, fixture.TfOptions, "workspace", "delete", workspace_name)
+	workspaceName := "unit-tst-" + random.UniqueId()
+	terraform.WorkspaceSelectOrNew(fixture.GoTest, fixture.TfOptions, workspaceName)
```
erikschlegel: if we're creating one workspace for each one of our parallel test runs, then wouldn't that conflict with multiple calls to `RunUnitTests`?

test-harness/infratests/integration.go
```diff
@@ -0,0 +1,117 @@
...
+//	- Run `terraform apply`
+//	- Run `terraform output`
+//	- Validate outputs
+//	- Run user-supplied validation of outputs
+//	- Destroy resources
+func RunIntegrationTests(fixture *IntegrationTestFixture) {
+	terraform.Init(fixture.GoTest, fixture.TfOptions)
+
+	workspaceName := "int-tst-" + random.UniqueId()
+	terraform.WorkspaceSelectOrNew(fixture.GoTest, fixture.TfOptions, workspaceName)
```
erikschlegel: Same comment from `RunUnitTests` applies here.

test-harness/infratests/unit.go
nmiodice: Good call. I suppose we want to limit the max number of terraform-based tests to 1 per host per working directory. Does this match your thinking?

erikschlegel: we could do that or support ♾️ test runs per working directory and use the same terraform workspace. I don't know the shortcomings of workspace's well enough to fully understand the risks with this approach. The other alternative is only supporting sequential test runs per working directory.

nmiodice: My understanding is that terraform operations over the same workspace (i.e., state file) should be limited

https://learn.hashicorp.com/terraform/development/running-terraform-in-automation
> It is strongly recommended to use a backend that supports remote state, since that allows Terraform to automatically save the state in a persistent location where it can be found and updated by subsequent runs. Selecting a backend that supports state locking will additionally provide safety against race conditions that can be caused by concurrent Terraform runs.

nmiodice: I suppose we can consider running each test deployment under the following condition:
 - Isolated temp directory: this lets us use a different workspace for each test. The state for this is kept, i believe, in `.terraform/` so we want a unique directory per test.
 - Unique workspace name (already done): this lets us run multiple deployments at once

nmiodice: In any case, I'd like to go ahead and punt this to a different user story as it is outside the scope of integration testing only.

I've created https://github.com/microsoft/cobalt/issues/141 to track this work

Issue #107, opened 2019-05-16
local-run.sh getopt check fails on ZSH


ericpera-msft:
getopt for OSX does not operate as expected in the ./test-harness/local-run.sh script. Below is an example of me trying to run the script from my MacBook Pro terminal:

> /test-harness/local-run.sh
>  --
> 'getopt --test' failed in this environment. please install getopt.
> install getopt using brew? [y,n]y
> ==> This script will install:
> /usr/local/bin/brew
> /usr/local/share/doc/homebrew
> /usr/local/share/man/man1/brew.1
> /usr/local/share/zsh/site-functions/_brew
> /usr/local/etc/bash_completion.d/brew
> /usr/local/Homebrew
> ==> Downloading and installing Homebrew...
> HEAD is now at 5a63de022 Merge pull request #6139 from Homebrew/dependabot/bundler/Library/Homebrew/backports-3.15.0
> Already up-to-date.
> ==> Installation successful!
> 
> ==> Homebrew has enabled anonymous aggregate formulae and cask analytics.
> Read the analytics documentation (and how to opt-out) here:
>   https://docs.brew.sh/Analytics
> 
> ==> Homebrew is run entirely by unpaid volunteers. Please consider donating:
>   https://github.com/Homebrew/brew#donations
> ==> Next steps:
> - Run `brew help` to get started
> - Further documentation:
>     https://docs.brew.sh
> Warning: gnu-getopt 2.33.2 is already installed and up-to-date
> To reinstall 2.33.2, run `brew reinstall gnu-getopt`
> you'll need to restart the shell instance to load the new path

-- Started a New Session -- 

> ./test-harness/local-run.sh
>  --
> 'getopt --test' failed in this environment. please install getopt.
> install getopt using brew? [y,n]

The getopt check fails even though getopt was just installed. If I run some getopt commands, this is the output:

> bash-5.0$ getopt --test
>  --
> bash-5.0$ echo "$?"
> 0
> bash-5.0$ getopt --test
>  --
> bash-5.0$ echo $PIPESTATUS
> 0

If we look at the man-page for getopt (http://man7.org/linux/man-pages/man1/getopt.1.html), we see that the GETOPT_COMPATIBLE environment variable might play a part in this. To verify that this variable is not set, here is output from me unsetting the variable before running the script:

>  unset GETOPT_COMPATIBLE
>  ./test-harness/local-run.sh
>  --
> 'getopt --test' failed in this environment. please install getopt.
[...]

Issue #106, opened 2019-05-16
Bash Path on OSX


ericpera-msft:
Line 1 in ./test-harness/local-run.sh (#!/bin/bash) will cause issues with OSX users that use homebrew to update Bash.

The default version of Bash located at /bin/bash is 3.4 on an OSX machine, which is incompatible with several commands in the ./test-harness/init.sh script. If OSX users use Homebrew to install an updated version of Bash, that installation is located at /usr/local/bin/bash and not at /bin/bash.

ericpera-msft:
Merged fix into master.

PR #130, opened 2019-05-22
Erperall/bugfix/fixes for osx

ericpera-msft:
## All Submissions:

Issue Number: #106, #107 

## What is the new behavior?

- Shebang loads bash interpreter in respective location for Linux and OSX.
- The bash profile is now loaded properly for OSX users to detect getopt installations.

- [NO]


test-harness/build-base-image.sh
```diff
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/usr/bin/env bash
+#/usr⁩/local⁩/⁨Cellar⁩/gnu-getopt⁩/2.33.2⁩/bin⁩/getopt
```
nmiodice: Is this line doing anything under the hood? To my untrained bash eye, it seems to be a comment only. Is there some pre-procssing magic that happens here?

If not, is it safe to remove?

test-harness/build-base-image.sh
```diff
@@ -23,6 +24,11 @@ function echoIfVerbose() {
         echo "${@}"
     fi
 }
+
+if [ -f ~/.bash_profile ]; then
+  . ~/.bash_profile
```
nmiodice: Should we be relying on the user to have executed this in their shell prior to invoking the script? I believe (correct me if I'm wrong here) that if they do this, the environment set in `.bash_profile` is inherited by the child process running the scripts.

test-harness/build-base-image.sh
ericpera-msft: Yup, that line was a mistake that should not have made it into the PR. Will remove it.

test-harness/build-base-image.sh
ericpera-msft: I've altered how this operates. Instead of loading the bash profile, the gnu-getopt installation location is loaded into a variable only within the scope of the script's execution.

ericpera-msft:
> Wondering if we should be making these changes in BashWizard, as I believe that they files were mostly generated from there.
> 
> Thoughts?

I believe these changes should be made to BashWizard if we want the generated bash script to be runnable by OSX. I'll reach out to Joe to discuss the proposed change with him.

frodopwns:
Is there any reason these scripts would ever be regenerated? I would say generated files should generally not be edited. Scaffolded files on the other hand, can probably be edited.

If BashWiz can't do OSX scripts, an issue should be opened and this PR should be considered for merging.

nmiodice:
Change looks good to me but we should bottom out on if this should be done in BashWizard before we decide to commit here or in that repo

ericpera-msft:
@nmiodice Talking to Joe Long now to determine the best course of action on that.

frodopwns:
We should fix it here AND in BashWizrd imo. Would help unblock people using OSX and `zsh` in the short term.

test-harness/local-run.sh
```diff
@@ -1,9 +1,11 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #---------- see https://github.com/joelong01/BashWizard ----------------
 # bashWizard version 1.0.0
 # this will make the error text stand out in red - if you are looking at these errors/warnings in the log file
 # you can use cat <logFile> to see the text in color.
 
+export FLAGS_GETOPT_CMD="$(brew --prefix gnu-getopt)/bin/getopt"
```
nmiodice:
```bash
$ export FLAGS_GETOPT_CMD="$(brew --prefix gnu-getopt)/bin/getopt"

Command 'brew' not found, but can be installed with:

sudo apt install linuxbrew-wrapper
```

Can we wrap this in an OS check?
What happens if the user does not have brew?

ericpera-msft: If the user does not have brew, the brew install for gnu-getopt would fail as well. Was going to bring this up to Joe to see if we should include an OS or dependency check in BashWizard, then propagate that here. As of right now, this PR is a WIP until a direction for the Wizard is settled on.

PR #181, opened 2019-06-17
Sets defaults for service principle module back to original

iphilpot:
## All Submissions:

During integration with the storage module, we found some issues with the default value for create_for_rbac being set to true. This value should be set to false to provide correct behavior. Also, a couple of the old names from variables carried over so I fixed them also.

Issue Number: N/A

## What is the new behavior?
create_for_rbac is false by default.

- [NO]

## Other information

### DIFF
--- a/infra/modules/providers/azure/service-principal/variables.tf
+++ b/infra/modules/providers/azure/service-principal/variables.tf
@@ -1,7 +1,7 @@
 variable "create_for_rbac" {
   description = "Create a new Service Principle"
   type        = bool
-  default     = true
+  default     = false
 }
 
 variable "display_name" {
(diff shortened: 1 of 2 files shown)


infra/modules/providers/azure/service-principal/variables.tf
```diff
@@ -1,7 +1,7 @@
 variable "create_for_rbac" {
   description = "Create a new Service Principle"
   type        = bool
-  default     = true
+  default     = false
 }
 
 variable "display_name" {
```
TechnicallyWilliams: Can we give this a default name? ex. ("RBAC Service Principle") Right now, clients using the module are forced to provide a "display name" even if it won't be used.

iphilpot: ok I'd file a bug on this.

Issue #169, opened 2019-06-10
Terraform Storage Account Module


erikschlegel:
### Description
As a technical operator, I'd like to deploy a storage container with custom defined access roles. System identity should be able to access the container without using a storage key.

### Acceptance Criteria
- [x] Automate the creation of a storage account and storage container.
- [x] Assign roles to the storage container.
- [x] Create a managed identity (MSI) from the storage container.
- [x] Export the principal id's of MSI and id of the storage container.

Also, here are a few points that need to be addressed:
1. Move to Terraform 12 and PR comments have reopened a conversation about implementing a storage account module that supports multi-container deployment.

### Tasks
- Decide on default configurations for both the storage account and storage container.
- Decide if scalar module (count vs a single instance module).
- Create Readme.md
- Re-discuss if storage account module supports single container or multi-container.

Issue #182, opened 2019-06-17
Service Principal module requires default name for existing sp


iphilpot:
Should add default name so it's optional?

```hcl
module "service-principal" {
  source                         = "../../modules/providers/azure/service-principal"
  object_id                      = "000000-0000-000-0000-000000"
 **display_name                   = "TFTester"
  role_scope                     = "/subscriptions/0b1f6471-1bf0-4dda-aec3-111122223333"
  role_name                      = "reader"
}
```

PR #183, opened 2019-06-18
Add an Azure storage module for managing storage accounts

TechnicallyWilliams:
## All Submissions:

Issue Number: #169 

## What is the new behavior?

Created a new module for deploying an Azure Storage Account along with an Azure Storage Container.
Module automatically enrolls the deployed Azure Storage Account into Managed Identities. 

In addition, module expects to receive a service principal id (the id of an Azure Resource/Service) for the purposes of assigning client access to the Azure Storage Account.

- [NO]

## Other information

App_to_Storage_RBAC


infra/modules/providers/azure/storage-account/README.md
```diff
@@ -0,0 +1,118 @@
+# Azure Storage
+
+This Terraform based `storage-account` module grants templates the ability to configure and deploy cloud `storage containers` along with a `storage account` using Microsoft's _**Azure Storage**_ service.
+
+In addition, this module offers both "authentication" and "authorization" features:
+
+- For "authentication", this module automatically enrolls the deployed `storage account` into Microsoft's _**Managed Identities**_ service.
```
nmiodice: [ nit ] adding quotes around `authentication` and `authorization` makes it sound dubious :D

infra/modules/providers/azure/storage-account/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+- For "authorization", this module expects to receive the `object id` of any Azure Resource that will behave as a client of the `storage-account`. Enrollment as an Azure Resource client is achieved through Microsoft's _**Role Based Access Secruity**_ (RBAC) service.
+
+#### _More on Azure Storage_
+
+> "A storage account provides a unique namespace in Azure for your data. Every object that you store in Azure Storage has an address that includes your unique account name. A container organizes a set of blobs, similar to a directory in a file system. A storage account can include an unlimited number of containers, and a container can store an unlimited number of blobs." - Source: Microsoft's [Introduction to Azure Blob Storage](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction)
+
+This module deploys a `storage account` along with a `storage container` in order to satisfy blob storage scenarios which are optimized for storing massive amounts of unstructured data, such as text or binary data.
```
nmiodice: Are we intending to support multiple containers in the account? Would this be achieved by specifying the resource you defined in this module multiple times with the same account name but different container names?

infra/modules/providers/azure/storage-account/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+Terraform resources directly referenced within the `storage-account` module include the following:
+
+- [azurerm_storage_account](https://www.terraform.io/docs/providers/azurerm/r/storage_account.html)
+
+- [azurerm_storage_container](https://www.terraform.io/docs/providers/azurerm/r/storage_container.html)
+
+## Usage
+
+App Monitoring usage example:
```
nmiodice: Wrong module name here

infra/modules/providers/azure/storage-account/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+- Ability to deploy a Storage Account automatically enrolled into _**Managed Identities**_ within a single resource group.
+
+- Ability to deploy a Storage Account with a role assigned Azure Resource client.
+
+- Ability to deploy a Storage Container within the Storage Account.
+
+## Definition
+
+Azure Storage definition example:
```
nmiodice: Is this the internal implementation of your module? It can be removed from the readme if it is.

infra/modules/providers/azure/storage-account/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+Please visit _**Azure Storage's**_ [Introduction to Azure Storage](https://docs.microsoft.com/en-us/azure/storage/common/storage-introduction.) page for more information on configuring a storage account.
+
+#### Configuring `rbac role`
+
+Please visit _**Cobalt's**_ [service principal module](../service-principal) page for more information on configuring a service principal and rbac role.
+
+## Argument Reference
+
+Supported arguments for this module are available in [variables.tf](variables.tf).
```
nmiodice: IMO this would be great to call out in the docs.

infra/modules/providers/azure/storage-account/main.tf
```diff
@@ -0,0 +1,32 @@
...
+  name                  = var.storage_container_name
+  resource_group_name   = var.resource_group_name
+  storage_account_name  = azurerm_storage_account.sa.name
+  container_access_type = "private"
+}
+
+module "service-principal" {
+  source       = "../service-principal"
+  object_id    = var.existing_sp_object_id
+  display_name = ""
```
nmiodice: Why `""`? Unclear on where this will pop up in the UI, but I can see having a display name being useful for tracking

infra/modules/providers/azure/storage-account/output.tf
```diff
@@ -0,0 +1,25 @@
+
+output "storage_account_id" {
+  description = "The ID of the storage account."
```
nmiodice: Is it possible to ad types to all of these? (note, I'm not 100% sure if types are supported in TF outputs)

infra/modules/providers/azure/storage-account/output.tf
```diff
@@ -0,0 +1,25 @@
...
+}
+
+output "storage_container_id" {
+  description = "The ID of the storage container from the storage account module."
+  value       = azurerm_storage_container.sa.id
+}
+
+output "storage_container_properties" {
+  description = "Map of additional properties associated with the storage container."
+  value       = azurerm_storage_container.sa.properties
```
nmiodice: What would these look like? Is there documentation for what to expect?

infra/modules/providers/azure/storage-account/variables.tf
```diff
@@ -0,0 +1,62 @@
...
+}
+
+variable "storage_container_name" {
+  description = "The name of the storage container. Must be unique under the storage account."
+  type        = "string"
+}
+
+variable "performance_tier" {
+  description = "Determines the level of performance required."
+  type        = "string"
```
nmiodice: FYI, `type = string` is the TF 12 way to do this, though `type = "string"` also works. More syntactically TF-12 though :)

infra/modules/providers/azure/storage-account/output.tf
TechnicallyWilliams: They are not supported.

image

infra/modules/providers/azure/storage-account/output.tf
TechnicallyWilliams: The ##Attributes Reference section at the bottom of the readme now has a code example for the keys that come with this map.

infra/modules/providers/azure/storage-account/main.tf
TechnicallyWilliams: This has been filed as a bug within our repo. #182  The "display name" var isn't used when taking the "existing service principal" path (the path my module has chosen). 

The "display name" var should only be required and accessed during "new service principal" scenarios. In other words, we desire that this variable be 'null' and not evaluated for "existing service principal" scenarios, however, it's evaluated even when null which causes an error. In order to avoid the error, i have to give it a value.

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: I'm more in favor of keeping the docs very light-weight. I think there's more value in incentivizing a reader to become familiar with the variables.tf as it's a very straight forward terraform file.

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: Quotes were removed in the latest commit.

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: This was updated in the latest commit.

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: You're right. There's no value add for having the implementation within the readme. This was removed in the latest commit.

infra/modules/providers/azure/storage-account/main.tf
nmiodice: oh interesting! thanks for pointing this out. Another option is to fix the bug by adding a default variable, seems like it would be simple to do and would drive down tech debt for our project.

Your call.

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: After our discussion, we were able to achieve multiple containers with different names.
multiple-containers

infra/modules/providers/azure/storage-account/variables.tf
TechnicallyWilliams: Yeah, we'll do it the TF 12 way. Thanks!

infra/modules/providers/azure/storage-account/main.tf
TechnicallyWilliams: Bug fix has been applied and merged into master.

infra/modules/providers/azure/storage-account/README.md
nmiodice: Containers*

infra/modules/providers/azure/storage-account/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Characteristics
+
+An instance of the `storage-account` module deploys the _**Azure Storage**_ service in order to provide templates with the following:
+
+- Ability to deploy a Storage Account automatically enrolled into _**Managed Identities**_ within a single resource group.
+
+- Ability to deploy a Storage Account with a role assigned Azure Resource client.
+
+- Ability to deploy a Storage Container within the Storage Account.
```
nmiodice: Containers*

infra/modules/providers/azure/storage-account/README.md
```diff
@@ -0,0 +1,96 @@
...
+  service_plan_name   = module.service_plan.service_plan_name
+  ...
+}
+
+module "storage_account" {
+  source                    = "../../modules/providers/azure/storage-account"
+  resource_group_name       = azurerm_resource_group.main.name
+  resource_group_location   = azurerm_resource_group.main.location
+  account_name              = "mystorageaccount"
+  storage_container_names   = ["astoragcontainer", "astoragecontainer"]
```
nmiodice: Why would I specify 2 containers with the same name?

infra/modules/providers/azure/storage-account/main.tf
```diff
@@ -0,0 +1,32 @@
...
+
+  # enrolls storage account into azure 'managed identities' authentication
+  identity {
+    type = "SystemAssigned"
+  }
+}
+
+resource "azurerm_storage_container" "sa" {
+  count                 = length(var.storage_container_names)
+  name                  = "${var.storage_container_names[count.index]}-${count.index}"
```
nmiodice: I think we should keep the name that the client passed in without adding the index. This way they have full control over the naming of their app dependencies

infra/modules/providers/azure/storage-account/variables.tf
```diff
@@ -0,0 +1,62 @@
...
+
+variable "kind" {
+  description = "Storage account types that determine available features and pricing of Azure Storage. Use StorageV2 when possible."
+  type        = string
+  default     = "StorageV2"
+}
+
+variable "https" {
+  description = "Boolean flag which forces HTTPS in order to ensure secure connections."
+  type        = string
```
nmiodice: Should this be typed as bool?

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: Updated container names in screenshot:
multiple-containers

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: updated to containers (plural)

infra/modules/providers/azure/storage-account/README.md
TechnicallyWilliams: Container names have been changed to be more example friendly. [ "movies", "photos"].

infra/modules/providers/azure/storage-account/main.tf
TechnicallyWilliams: Makes sense! Storage account module no longer appends index count to container names. Thanks!

infra/modules/providers/azure/storage-account/variables.tf
TechnicallyWilliams: Thanks for pointing this out! This variable now uses bool as a type.

PR #185, opened 2019-06-20
Service Principal Default Name Bug

TechnicallyWilliams:
## All Submissions:

Issue Number: #182 

## What is the new behavior?

Service Principal module now has a default name for 'display name' variable when creating new service principals. 

Users of the module passing in 'existing service principals' no longer have to pass in a value intended for 'new service principal' scenarios. .

- [NO]

Required vars have been moved to the top of the script.

## Other information

image

### DIFF
--- a/infra/modules/providers/azure/service-principal/variables.tf
+++ b/infra/modules/providers/azure/service-principal/variables.tf
@@ -1,20 +1,3 @@
-variable "create_for_rbac" {
-  description = "Create a new Service Principle"
-  type        = bool
-  default     = false
-}
-
-variable "display_name" {
-  description = "Display name of the AD application"
-  type        = string
-}
-
-variable "object_id" {
-  description = "Object Id of the service principle to assign a role"
-  type        = string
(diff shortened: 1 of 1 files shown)


infra/modules/providers/azure/service-principal/variables.tf
```diff
@@ -7,21 +17,11 @@ variable "create_for_rbac" {
 variable "display_name" {
   description = "Display name of the AD application"
   type        = string
+  default     = "AD Client"
```
iphilpot:
```suggestion
  default     = ""
```

I would leave this blank because it's what is expected if creating a role assignment using an existing SP. Also if blank when creating a new AAD App then a name will be generated.

TechnicallyWilliams: Updated.

nmiodice: +1. Other than that this LGTM

Issue #188, opened 2019-06-24
Implement "subscription 1" components of isolated service template


nmiodice:
### Description
*As an enterprise customer I'd like to deploy an app service into an ASE so that I can verify that I am running in a secure VNET*

### Acceptance Criteria
- [x] 2 app services should be deployed to an ASE instance (with unit/integ tests). These should be deployed to a new resource group + service plan
- [x] The app service should be enabled for CD. However, as we won't be creating an ACR we don't need to configure the CD webhooks
- [x] An app insights instance should be deployed alongside the app service (with unit/integ tests)

### Resources
 - Technical Design: https://github.com/microsoft/cobalt/blob/master/infra/templates/az-isolated-service-single-region/docs/design.md

### Tasks
 - [x] Create VNET for testing. The VNET
 - [x] Create ASE for testing. The ASE + VNET should have an outwardly accessible port that enables testing
 - [x] Implement template by composing existing modules
 - [x] Implement unit tests for template
 - [x] Implement integration tests for template
 - [x] PR
 - [ ] Demo

PR #194, opened 2019-06-26
Implement admin components of az-isolated-service-single-region template.

nmiodice:
## All Submissions:

No template exists for `az-isolated-service-single-region`

Issue Number: https://github.com/microsoft/cobalt/issues/188

## What is the new behavior?
`az-isolated-service-single-region` template partially implemented. The template should include the resources called out in [the design](https://github.com/microsoft/cobalt/blob/master/infra/templates/az-isolated-service-single-region/docs/design.md) that are deployed to subscription #1.

- [NO]

 - 2 Deployed Apps w/ Staging Slots
apps

- App Service Plan deployed to ASE
asp

## Notes
 - There is a [follow on PR](https://github.com/terraform-providers/terraform-provider-azurerm/pull/3737) to handle the scenario where `Isolated` app services can take > 1 hour to deploy


devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -48,6 +48,10 @@ variables:
   value: 'xterm-color'
 - name: TF_WARN_OUTPUT_ERRORS
   value: 1
+- name: STATIC_ASE_NAME
```
erikschlegel: Should we move this into the pipeline variable settings within the azure devops portal as opposed to defining it within our build definition?

infra/templates/az-isolated-service-single-region/.env.template
```diff
@@ -0,0 +1,15 @@
...
+TF_VAR_remote_state_account="<tf-remote-state-storage-account-name>"
+TF_VAR_remote_state_container="<tf-remote-state-storage-container-name>"
+TF_VAR_admin_subscription_id="<subscription for hosting admin resources>"
+TF_VAR_app_service_environment_name="<name of app service environment>"
+TF_VAR_app_service_environment_resource_group="<resource group hosting app service environment>"
+GO_VERSION=1.12.5
+TF_VERSION=0.12.2
+BUILD_BUILDID=1
+DATACENTER_LOCATION=eastus2
+TF_WARN_OUTPUT_ERRORS=1
```
erikschlegel: Wouldn't this mean that we'd have to redefine the `.env` for each template?

nmiodice: Yes it does. For example, an AWS template will need different environment variables from an Azure template

devops/providers/azure-devops/templates/azure-pipelines.yml
nmiodice: Honestly I'd love some guidance on this one -- this choice has always seemed pretty arbitrary to me. I suppose the pros of putting it into AZDO is that other users of this could define it differently.

Aside from this, is there guidance for deciding where to put these variables?

infra/templates/az-isolated-service-single-region/.env.template
nmiodice: removed.

devops/providers/azure-devops/templates/azure-pipelines.yml
nmiodice: I converted these to pipeline variables.

infra/templates/az-isolated-service-single-region/.env.template
erikschlegel: Concerned about the developer experience considering the scenario we have 20+ cobalt templates for azure where most of those `.env` files will comprise of duplicate environment definitions. An alternate approach would be to support `.env ` overlays in each template folders but still have the root `.env` file provide the default environment.

infra/templates/az-isolated-service-single-region/terraform.tfvars
```diff
@@ -0,0 +1,9 @@
+# GENERAL
+resource_group_location = "eastus2"
+name                    = "isolated-service"
+
+# SERVICE PLAN
+app_service_name = {
+  cobalt-backend-api-1 = "appsvcsample/static-site:latest"
```
erikschlegel: I'd imagine this definition of `app_service_name` would only be used for local testing. 

As part of our provisioning process, we should build a fresh image for each app service, similar to our single region template. The approach we took with the single region template was referencing the `https://github.com/erikschlegel/echo-server.git` image as a placeholder. The advantage is that we can tag each image to a release that's somewhat specific to the workspace ie `msftcse/cobalt-backend-api-1:release-devint`, and avoid creating a webhook on an image that's tagged to `latest`. 

```
app_service_name = {
    cobalt-backend-api = "msftcse/az-service-single-region:release" 
}
# ACR image deploy source
acr_build_git_source_url = "https://github.com/erikschlegel/echo-server.git"
```

The idea here is that app devs provide the image repo name and tag per app service, and we create a placeholder image. 

In the azure devops pipeline, `app_service_name` and `acr_build_git_source_url` will be provided as pipeline variables.

infra/templates/az-isolated-service-single-region/admin.tf
```diff
@@ -0,0 +1,63 @@
...
+  app_service_environment_id = local.ase_id
+  providers = {
+    "azurerm" = "azurerm.admin"
+  }
+}
+
+module "app_service" {
+  source                           = "../../modules/providers/azure/app-service"
+  app_service_name                 = var.app_service_name
+  service_plan_name                = module.service_plan.service_plan_name
```
erikschlegel: How are the docker credentials being specified?

infra/templates/az-isolated-service-single-region/admin.tf
```diff
@@ -0,0 +1,63 @@
...
+  service_plan_resource_group_name = azurerm_resource_group.rg.name
```
erikschlegel: We're missing the `vault_uri` setting

infra/templates/az-isolated-service-single-region/admin.tf
```diff
@@ -0,0 +1,63 @@
...
+  service_plan_size          = var.service_plan_size
+  service_plan_kind          = var.service_plan_kind
+  app_service_environment_id = local.ase_id
+  providers = {
+    "azurerm" = "azurerm.admin"
+  }
+}
+
+module "app_service" {
+  source                           = "../../modules/providers/azure/app-service"
```
erikschlegel: can you confirm that we don't need to specify the `vnet_name` and `vnet_subnet_id` to enable vnet isolation for ASE?

infra/templates/az-isolated-service-single-region/terraform.tfvars
```diff
@@ -0,0 +1,9 @@
+# GENERAL
+resource_group_location = "eastus2"
```
erikschlegel: Can we confirm if this region is the same that our target customer is using in their ASE envrionment?

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// to fail for unrelated reasons.
+	testFixture := infratests.IntegrationTestFixture{
+		GoTest:                t,
+		TfOptions:             tfOptions,
+		Workspace:             workspace,
+		SkipCleanupAfterTest:  true,
+		ExpectedTfOutputCount: 1,
+		ExpectedTfOutput: infratests.TerraformOutput{
+			"fqdns": []string{
+				"http://cobalt-backend-api-1-" + workspace + "." + ase_name + ".p.azurewebsites.net",
```
erikschlegel: I'd imagine that only https connections will be allowed in an enterprise environment.

infra/templates/az-isolated-service-single-region/variables.tf
```diff
@@ -0,0 +1,140 @@
...
+  description = "The name key value pair where the key is the name assigned to the app service and value is the source container"
+  type        = map(string)
+}
+
+
+// ---- Service Plan Configuration ----
+
+variable "scaling_values" {
+  description = "Dimensions used to determine service plan scaling"
+  type        = list(string)
```
erikschlegel: Naive question. Does terraform 12 happen to support strong typed collections?

erikschlegel: Should we use strong typed collections similar to `scaling_rules`?

infra/templates/az-isolated-service-single-region/variables.tf
```diff
@@ -0,0 +1,140 @@
...
+
+variable "app_service_name" {
+  description = "The name key value pair where the key is the name assigned to the app service and value is the source container"
+  type        = map(string)
+}
+
+
+// ---- Service Plan Configuration ----
+
+variable "scaling_values" {
```
erikschlegel: We should rename this to `monitoring_dimension_values` so the naming convention is more accurate to it's use.

infra/templates/az-isolated-service-single-region/terraform.tfvars
nmiodice: The deployment will fail if they don't match so this validation is done server side for us.

There is no data resource available for us to load this up natively so I'm ok leveraging server side validation

infra/templates/az-isolated-service-single-region/admin.tf
nmiodice: Oops, I'm not passing them in...

infra/templates/az-isolated-service-single-region/admin.tf
nmiodice: This resource is created in another subscription which is out of scope for the story I'm working on.

We need follow ups for key vault along with the other resources created there, in addition to configuring permissions etc...

infra/templates/az-isolated-service-single-region/terraform.tfvars
nmiodice: See note about resources from the other subscription being out of scope. Based on this I think that we may want to put this off until ACR integration

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
nmiodice: Key vault is out of scope for this story so I was planning on leaving https/certs till we pick up that work

infra/templates/az-isolated-service-single-region/.env.template
nmiodice: I'm going to delete this file for now since the only additions are `TF_VAR_*` which need not be environment-specific. 

I agree that we should think more about this before landing on a new approach.

infra/templates/az-isolated-service-single-region/variables.tf
```diff
@@ -0,0 +1,140 @@
+// ---- General Configuration ----
+
+variable "admin_subscription_id" {
```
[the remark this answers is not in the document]
nmiodice: Done

infra/templates/az-isolated-service-single-region/admin.tf
nmiodice: Confirmed.

Your next question may be "then how do we test this...?" which is a good question. For some context, I've configured the ASE we are using to enable external access through a single virtual IP address. Documentation here: https://docs.microsoft.com/en-us/azure/app-service/environment/network-info

Here are how the IP addresses work. note they all use the same IP address (managed by the VIP in ASE):

| IP ADDRESS | SUB-DOMAIN | DOMAIN |
|---|---|---|
| 52.177.217.27 | cobalt-backend-api-1-cobalt-isolated-testing(staging) | cobalt-static-ase.p.azurewebsites.net |
| 52.177.217.27 | cobalt-backend-api-1-cobalt-isolated-testing | cobalt-static-ase.p.azurewebsites.net |
| 52.177.217.27 | cobalt-backend-api-2-cobalt-isolated-testing | cobalt-static-ase.p.azurewebsites.net |
| 52.177.217.27 | cobalt-backend-api-2-cobalt-isolated-testing(staging) | cobalt-static-ase.p.azurewebsites.net |

If a customer does not want this, the configuration is at the ASE level, not the app-service or app-service plan level.

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
nmiodice: Based on our discussion we will focus on this when we pick up the key-vault integration work as a part of https://github.com/microsoft/cobalt/issues/195

infra/templates/az-isolated-service-single-region/admin.tf
nmiodice: This will be integrated as a part of https://github.com/microsoft/cobalt/issues/195 based on the discussion we had earlier today.

infra/templates/az-isolated-service-single-region/terraform.tfvars
nmiodice: Based on our convo this will be done when we implement ACR integration for the template as outlined in https://github.com/microsoft/cobalt/issues/196

infra/templates/az-isolated-service-single-region/admin.tf
nmiodice: Based on our convo this will be done when we implement ACR integration for the template as outlined in https://github.com/microsoft/cobalt/issues/196

infra/modules/providers/azure/app-monitoring/variables.tf
```diff
@@ -63,7 +63,7 @@ variable "metric_alert_criteria_threshold" {
   type        = string
 }
 
-variable "scaling_values" {
+variable "monitoring_dimension_values" {
   description = "App instance names made available from app service plan scaling options."
```
iphilpot: Nit: I don't know what this is from the description given. Can we explain this better?

infra/templates/az-isolated-service-single-region/docs/design.md
```diff
@@ -85,7 +85,7 @@ Only one terraform template can be deployed within a given workspace. The use ca
 | `name` | string |  | The name of the deployment.  This will be used across the resource created in this solution |
 | `app_service_containers` | map(str) |  | The name key value pair where the key is the name assigned to the app service and value is the source container |
 | `metrics_configuration` | MetricsConfig **see below** | `[]` | A list of metrics configuration that should be configured for the app service plan |
-| `scaling_values` | `["*"]` |  | Dimensions used to determine service plan scaling |
+| `monitoring_dimension_values` | `["*"]` |  | Dimensions used to determine service plan scaling |
```
iphilpot: this is a much more leading description that in the variable file

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+		"ase_subscription_id":     admin_subscription,
+		"ase_name":                ase_name,
+		"ase_resource_group":      ase_resource_group,
+		"app_service_name": map[string]interface{}{
+			"cobalt-backend-api-1": "appsvcsample/static-site:latest",
+			"cobalt-backend-api-2": "appsvcsample/static-site:latest",
+		},
+	},
+	BackendConfig: map[string]interface{}{
+		"storage_account_name": os.Getenv("TF_VAR_remote_state_account"),
```
iphilpot: Suggestion: may want to look at envy for working with envvars. Adds a lot of functionality and error handling.

https://github.com/gobuffalo/envy

infra/templates/az-isolated-service-single-region/tests/unit/unit_test.go
```diff
@@ -0,0 +1,149 @@
...
+			"module.app_insight.azurerm_application_insights.appinsights":                  expectedAppInsights,
+			"module.app_service.azurerm_app_service.appsvc[0]":                             expectedAppService1,
+			"module.app_service.azurerm_app_service.appsvc[1]":                             expectedAppService2,
+			"module.app_service.azurerm_app_service_slot.appsvc_staging_slot[0]":           expectedStagingSlot,
+			"module.app_service.azurerm_app_service_slot.appsvc_staging_slot[1]":           expectedStagingSlot,
+			"module.service_plan.azurerm_monitor_autoscale_setting.app_service_auto_scale": expectedAutoScalePlan,
+		},
+	}
+
+	infratests.RunUnitTests(&testFixture)
```
iphilpot: looks like a complete set of tests

test-harness/magefile.go
```diff
@@ -93,7 +93,7 @@ func verifyRunsQuietly(cmd string, args ...string) error {
 		return nil
 	}
 
-	return fmt.Errorf("ERROR: command '%s' with arguments %s failed. Output was: '%s'", cmd, args, output)
+	return fmt.Errorf("ERROR: command '%s' with arguments %s failed. Output was: '%s'. %s", cmd, args, output, instructionsToFix)
```
iphilpot: I like this. Looks like a very complete message that should lead users to understand what went wrong.

infra/templates/az-isolated-service-single-region/admin.tf
iphilpot: I too was curious how this works. Thanks for the explanation and doc.

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
nmiodice: Thanks for sharing this. We discussed that this is a good path moving forward but no need to switch at the moment.

infra/modules/providers/azure/app-monitoring/variables.tf
nmiodice: Clarified to `Dimensions used to determine service plan scaling.`

Issue #195, opened 2019-06-28
Implementation: Key Vault (w/ VNET) integration for ASE template


nmiodice:
### Description
*As a user of the `az-isolated-service-single-region` template, I'd like to leverage key vault integration in order to store secrets for my application.*

### Acceptance Criteria
- [x] Key Vault integration to `az-isolated-service-single-region` should be achieved. This is intended to be deployed into the `app dev` subscription, so the azure provider used to create the module should be configured to use that subscription.
- [x] There should be VNET isolation configured for the Key Vault instance
- [x] ~~Create backlog item for a better testing strategy for VNET isolated resources~~ Plan has been identified: write extensions for Terratest to inspect resources using the Azure API.

### Resources
Here is the technical design for this template. https://github.com/microsoft/cobalt/blob/master/infra/templates/az-isolated-service-single-region/docs/design.md

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

Issue #196, opened 2019-06-28
Implementation: ACR integration (including initial image build/pushes) for ASE template


nmiodice:
### Description
*As a user of the `az-isolated-service-single-region` template, I'd like a new ACR to be created and my web apps to have CD integrated with this ACR in order to more easily deploy to my applications.*

### Acceptance Criteria
- [x] ACR integration to `az-isolated-service-single-region` should be achieved. This is intended to be deployed into the `app dev` subscription, so the azure provider used to create the module should be configured to use that subscription.
- [x] The `app_service_name` variable should be renamed and refactored to include metadata required to configure CD for the app service. This should be a list of metadata blobs where each blob specifies things like (just examples, define correct schema during implementation) (1) Git repo to trigger initial build, (2) path to docker file, (3) image name, (4) image tag, etc...
- [x] The initial apply (*only the initial apply*) of the terraform template should trigger the images to be built and pushed to the ACR via ACR build tasks
- [x] There should be VNET isolation configured for the ACR. This will need to be done *after* the build task completes
- [x] There should be a webhook set up for each image built. This should trigger a deployment to the app service whenever an image is pushed to ACR.
- [x] The changes should be tested with unit and integration tests

### Resources
Here is the technical design for this template. https://github.com/microsoft/cobalt/blob/master/infra/templates/az-isolated-service-single-region/docs/design.md

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

PR #202, opened 2019-07-02
KeyVault with VNET integration + Terratest azure extension module

nmiodice:
## All Submissions:

No keyvault integration or VNET integration support for keyvault

Issue Number: https://github.com/microsoft/cobalt/issues/195

## What is the new behavior?
Keyvault integration with VNET integration

- [NO]


infra/templates/az-isolated-service-single-region/app.tf
```diff
@@ -0,0 +1,35 @@
+// This file contains all of the resources that exist within the app dev subscription. Design documentation
+// with more information on exactly what resources live here can be found at ./docs/design.md
+
+// Note: unfortunately the alias cannot be configured by passing a variable through
+// the module initialization!
+provider "azurerm" {
+  alias           = "app_dev"
+  subscription_id = var.app_dev_subscription_id
```
TechnicallyWilliams: Where does var.app_dev_subscription_id derive it's value? I'm not seeing any calls to the env anywhere.

infra/templates/az-isolated-service-single-region/ase.tf
```diff
@@ -55,8 +44,8 @@ module "app_service" {
   source                           = "../../modules/providers/azure/app-service"
   app_service_name                 = var.app_service_name
```
TechnicallyWilliams: Should this be app_service_name(s) instead? At first glance, it looks dedicated to a single application.

nmiodice: I agree. This came up in a previous PR as well but I want to wait to make any changes to this until we pick up https://github.com/microsoft/cobalt/issues/196. This variable will need to be renamed and refactored to have a richer schema that allows us to define things like (1) git repo, (2) location of docker file, (3) image name and (4) image tag (along with other metadata most likely)

Based on this I think it makes sense to defer this work until we pick up https://github.com/microsoft/cobalt/issues/196 (most likely next week)

infra/templates/az-isolated-service-single-region/app.tf
nmiodice: This is defined in `infra/templates/az-isolated-service-single-region/variables.tf`. The value will come by supplying a `-var` param when running `terraform plan|apply` or by declaring `TF_VAR_app_dev_subscription_id` as an environment variable, which is done in the build pipeline. The variable for the build pipeline is configured in AZDO as an environment variable.

I also have decided to use the env-var approach since its easier on my local. Its not in the `.env.template`, though, since it is not actually required as an environment variable.

TechnicallyWilliams: Is the intention command line only?

nmiodice: Not necessarily. There are a number of options that you may want to use depending on your deployment strategy. See below. I think that the implementation of this was not covered in any design thus far

https://www.terraform.io/docs/configuration/variables.html#assigning-values-to-root-module-variables

infra/templates/az-isolated-service-single-region/app.tf
```diff
@@ -0,0 +1,35 @@
+// This file contains all of the resources that exist within the app dev subscription. Design documentation
```
TechnicallyWilliams: Is this the file that's expected to grow as more and more resources need vnet integration?

nmiodice: yes -- we'll be adding things like ACR to this file as they all fall within the app-dev subscription.

infra/modules/providers/azure/keyvault/main.tf
```diff
@@ -27,6 +25,16 @@ resource "azurerm_key_vault" "keyvault" {
     certificate_permissions = var.keyvault_certificate_permissions
   }
 
+  # This block configures VNET integration if a subnet whitelist is specified
+  dynamic "network_acls" {
+    for_each = length(var.subnet_id_whitelist) == 0 ? [] : [var.subnet_id_whitelist]
+    content {
+      bypass         = "None"
+      default_action = "Deny"
```
iphilpot: Should this be "deny" if it's for creating whitelists?

infra/templates/az-isolated-service-single-region/app.tf
iphilpot: Have we put this in our docs for devs?

infra/templates/az-isolated-service-single-region/outputs.tf
```diff
@@ -1,6 +1,39 @@
...
+output "app_dev_resource_group" {
+  value = azurerm_resource_group.app_rg.name
+}
+
+output "keyvault_name" {
+  value = module.keyvault.keyvault_name
+}
+
+output "keyvault_id" {
+  value = module.keyvault.keyvault_id
```
iphilpot: Why do we need to surface this as an output?

infra/templates/az-isolated-service-single-region/query_subnet_vnet_ids.sh
```diff
@@ -0,0 +1,34 @@
...
+#   [
+#       { name: <subnet name 1>, id: <subnet ID 1> },
+#       { name: <subnet name 2>, id: <subnet ID 2> },
+#       ...
+#   ]
+SUBNET_LISTING=$(az network vnet subnet list \
+    --subscription $1       \
+    --resource-group $2     \
+    --vnet-name $3          \
+    --output json           \
```
iphilpot: Unless you need the JSON, you can use `--output tsv` which is tab-separated values, with no keys or simply "plain text". Should remove the dependency on jq...

test-harness/docker/base-images/Dockerfile
```diff
@@ -13,7 +13,7 @@
 ARG gover
 FROM golang:${gover}-stretch as build
 RUN apt-get update && \
-    apt-get install -y git curl apt-transport-https lsb-release gpg
+    apt-get install -y git curl apt-transport-https lsb-release gpg jq
```
iphilpot: might be able to remove jq if tsv works for you.

test-harness/terratest-extensions/modules/azure/authorizer.go
```diff
@@ -0,0 +1,29 @@
+package azure
```
iphilpot: I'd like to discuss the decision to put this in the Azure package... I believe this is going to be very similar to what we need to use for integration testing and would like a single location for this. I believe you'll know best where that is.

test-harness/terratest-extensions/modules/azure/network.go
```diff
@@ -0,0 +1,48 @@
...
+
+// GetSubnetsInVnetE - Return the subnets that exist wihin a given VNET
+func GetSubnetsInVnetE(subscriptionID string, resourceGroupName string, vnetName string) ([]string, error) {
+
+	client, err := getVirtualNetworkClient(subscriptionID)
+	if err != nil {
+		return nil, err
+	}
+
+	vnet, err := client.Get(context.Background(), resourceGroupName, vnetName, "")
```
iphilpot: I feel like we should pass in the context so that we can reuse the context throughout the tests. I also believe we could drop the bearer token in context.value and reuse it if needed. (I don't think we'll need to worry much about token timeout)

infra/templates/az-isolated-service-single-region/outputs.tf
```diff
@@ -1,6 +1,39 @@
...
+
+output "admin_resource_group" {
+  value = azurerm_resource_group.admin_rg.name
+}
+
 output "fqdns" {
   value = [
-      for uri in module.app_service.app_service_uri:
-        "http://${uri}"
+    for uri in module.app_service.app_service_uri :
```
erikschlegel: Yay Terraform 12!

infra/modules/providers/azure/keyvault/main.tf
nmiodice: As I understand you need to specify 2 things in order to restrict access to a, pardon my pun, subset of subnets. (1) is to specifically deny all traffic by default and (2) is to specifically accept traffic from the specified subnets

infra/templates/az-isolated-service-single-region/app.tf
nmiodice: No, there is no readme for this template yet but I think it would be a good idea to add it to the README when we start documentation.

Perhaps the documentation should already have been started but I was planning on writing it up after things normalize with the inputs/outputs and whatnot

infra/templates/az-isolated-service-single-region/outputs.tf
nmiodice: I thought i'd need it to access the keyvault resource when writing my tests but I think its actually not needed. I'll remove it.

infra/templates/az-isolated-service-single-region/query_subnet_vnet_ids.sh
nmiodice: Unfortunately terraform only accepts JSON from external programs. Even more of a pain is that it only accepts string keys and string values. This is why I am using `jq` to transform the json output.

I tried really hard to avoid this but was unable to. Happy for an alternative solution, though! LMK if anything comes to mind.

test-harness/terratest-extensions/modules/azure/authorizer.go
nmiodice: I put it here because the current implementation makes a lot of assumptions about the format of environment variables. Since this runs within the context of terraform, its safe to assume things like `ARM_CLIENT_ID`, `ARM_CLIENT_SECRET`, etc... exist since that is what terraform is looking for.

If we are going to make this generic, I think that it makes sense to parameterize the following:
- ARM_CLIENT_SECRET
- ARM_CLIENT_ID
- ARM_TENANT_ID
- OAuth endpoint (line 14)
- Token Endpoint (line 23)

Not sure where it would live but I agree that there is probably room for improvement here. Would like to discuss this offline w/ you.

test-harness/terratest-extensions/modules/azure/network.go
nmiodice: I'd love to understand this more :). I'll be the first to admid that I went with "first thing to work" on this portion of the PR and would love to learn how to improve it.

I'll ramp up on what these contexts are before picking your brain.

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
```diff
@@ -62,6 +66,42 @@ func httpGetRespondsWith200(goTest *testing.T, output infratests.TerraformOutput
 	}
 }
 
+// Verifies that the Key Vault instance deployed is properly isolated within the VNET
+func verifyVnetIntegrationForKeyVault(goTest *testing.T, output infratests.TerraformOutput) {
+	vaultResourceGroup := output["app_dev_resource_group"].(string)
+	vaultName := output["keyvault_name"].(string)
+	keyVaultACLs := azure.GetKeyVaultNetworkAcls(goTest, adminSubscription, vaultResourceGroup, vaultName)
+	subnetIDs := azure.GetSubnetsInVnet(goTest, adminSubscription, aseResourceGroup, os.Getenv("TF_VAR_ase_vnet_name"))
```
erikschlegel: as discussed, would be great to prefix the function name with the scoped resource type. For example `GetSubnetsInVnet` -> `VnetSubnetsList`,  `getVirtualNetworkClient` -> `vnetClient`

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
```diff
@@ -62,6 +66,42 @@ func httpGetRespondsWith200(goTest *testing.T, output infratests.TerraformOutput
...
+
+	// No azure services should have bypass rules that allow them to circumvent the VNET isolation
+	if keyVaultACLs.Bypass != keyvault.None {
+		goTest.Fatal(fmt.Errorf("Expected bypass option of %s but got %s", keyvault.None, keyVaultACLs.Bypass))
+	}
+
+	// The default action should be to deny all traffic
+	if keyVaultACLs.DefaultAction != keyvault.Deny {
```
erikschlegel: you may want to consider using `assert.Equal` here from `github.com/stretchr/testify/assert`.

infra/templates/az-isolated-service-single-region/tests/integration/integration_test.go
```diff
@@ -76,16 +116,32 @@ func TestAzureSimple(t *testing.T) {
...
 			httpGetRespondsWith200,
+			verifyVnetIntegrationForKeyVault,
 		},
 	}
+
+	azureLogin(t)
 	infratests.RunIntegrationTests(&testFixture)
 }
+
+func azureLogin(t *testing.T) {
```
erikschlegel: We should move this to `authorizer.go`

infra/templates/az-isolated-service-single-region/tests/unit/unit_test.go
```diff
@@ -145,5 +158,19 @@ func TestTemplate(t *testing.T) {
 		},
 	}
 
+	// Required because there is a VNET query done by the template that requires a call to Azure CLI
+	azureLogin(t)
 	infratests.RunUnitTests(&testFixture)
 }
+
+func azureLogin(t *testing.T) {
```
erikschlegel: let's centralize the login step into a single function managed within the azure extension module.

infra/templates/az-isolated-service-single-region/app.tf
nmiodice: I'll do this as part of https://github.com/microsoft/cobalt/issues/204

test-harness/terratest-extensions/modules/azure/authorizer.go
nmiodice: Refactored after discussing offline.

PR #205, opened 2019-07-10
ASE Template | ACR | CD Hooks | VNET Integration

nmiodice:
## All Submissions:
* [N/A] I have updated the documentation accordingly.

ASE Template missing ACR, CD Hook
Issue Number: https://github.com/microsoft/cobalt/issues/196

## What is the new behavior?
ACR w/ VNET integration + CD Hooks

- [NO]


test-harness/terratest-extensions/modules/azure/arc.go
```diff
@@ -0,0 +1,42 @@
+package azure
```
iphilpot: Need to change the name of this file from `arc.go` to `acr.go`

infra/modules/providers/azure/container-registry/output.tf
```diff
@@ -13,3 +13,12 @@ output "container_registry_name" {
   value       = var.container_registry_name
 }
 
+output "admin_username" {
+  description = "If admin access is enabled, this will be the username for the ACR"
+  value = azurerm_container_registry.container_registry.admin_username
+}
+
+output "admin_password" {
```
erikschlegel: We should avoid exporting sensitive values like passwords and private keys from our modules.

infra/templates/az-isolated-service-single-region/app.tf
```diff
@@ -33,3 +33,76 @@ module "keyvault" {
     "azurerm" = "azurerm.app_dev"
   }
 }
+
+module "container_registry" {
+  source                  = "../../modules/providers/azure/container-registry"
+  container_registry_name = local.acr_name
+  resource_group_name     = azurerm_resource_group.app_rg.name
+  // Note: this is requird until App Services and ACR work over MSI. See the design document for more details.
+  container_registry_admin_enabled = true
```
erikschlegel: This should be driven from a template variable as the value of admin_enabled will be changing soon.

infra/templates/az-isolated-service-single-region/app.tf
```diff
@@ -33,3 +33,76 @@ module "keyvault" {
...
+  // Note: only premium ACRs allow configuration of network access restrictions
+  container_registry_sku = "Premium"
```
erikschlegel: Why are we hardcoding `Premium` as the sku?

infra/templates/az-isolated-service-single-region/app.tf
```diff
@@ -33,3 +33,76 @@ module "keyvault" {
...
+# Build and push the app service image slot to enable continuous deployment scenarios.
+# We're using ACR build tasks to remotely carry out the docker build / push.
+resource "null_resource" "acr_image_deploy" {
+  count      = length(var.deployment_targets)
+  depends_on = [module.container_registry]
+
+  triggers = var.deployment_targets[count.index]
+
+  provisioner "local-exec" {
+    command = "az acr build -t $IMAGE_TAG -r $DOCKER_REPOSITORY -f $DOCKER_FILE $GIT_SOURCE"
```
erikschlegel: Have you explored what's involved with leveraging acr build tasks to source the personal access token for private hld repos?

infra/templates/az-isolated-service-single-region/app.tf
```diff
@@ -33,3 +33,76 @@ module "keyvault" {
...
+    //      are constrained to only allow access from within the configured VNET.
+    //  (2) A subsequent deployment specifies N + 1 webapps. This will attempt to trigger an ACR build task
+    //      that builds and pushes the new container to the ACR. However, this will fail due to the network
+    //      restrictions configured after the initial ACR build tasks ran.
+    //
+    // This flag will ignore the failed build from (2). The implication is that the webapp will deploy
+    // successfully. Hitting the HTTP(s) endpoint will return an error response because the container that
+    // should be deployed does not exist. The fix for this is to build and push the image to the ACR from
+    // within the VNET, which will be done as a part of the app dev CI/CD process.
+    on_failure = "continue"
```
erikschlegel: It's unclear how we'd bubble up errors if the initial docker build step were to fail?

infra/templates/az-isolated-service-single-region/variables.tf
```diff
@@ -38,9 +38,31 @@ variable "ase_vnet_name" {
...
+    image_name               = string
+    image_release_tag_prefix = string
+  }))
+  default = [
+    {
+      app_name                 = "cobalt-backend-api-1"
+      repository               = "https://github.com/erikschlegel/echo-server.git"
+      dockerfile               = "Dockerfile"
+      image_name               = "appsvcsample/static-site-1"
+      image_release_tag_prefix = "latest"
```
erikschlegel: this should be `release`

infra/templates/az-isolated-service-single-region/ase.tf
```diff
@@ -42,10 +42,17 @@ module "service_plan" {
...
 module "app_service" {
   source                           = "../../modules/providers/azure/app-service"
-  app_service_name                 = var.app_service_name
   service_plan_name                = module.service_plan.service_plan_name
   service_plan_resource_group_name = azurerm_resource_group.admin_rg.name
   app_insights_instrumentation_key = module.app_insights.app_insights_instrumentation_key
+  azure_container_registry_name    = module.container_registry.container_registry_name
+  docker_registry_server_url       = module.container_registry.container_registry_login_server
+  docker_registry_server_username  = module.container_registry.admin_username
+  docker_registry_server_password  = module.container_registry.admin_password
```
erikschlegel: as mentioned in my earlier note, I'm a bit concerned that we're pulling the acr password from terraform state as opposed to the terraform acr data source reference.

infra/templates/az-isolated-service-single-region/variables.tf
```diff
@@ -38,9 +38,31 @@ variable "ase_vnet_name" {
...
+variable "deployment_targets" {
+  description = "Metadata about apps to deploy, such as repository location, docker file metadata and image names"
+  type = list(object({
+    app_name                 = string
+    repository               = string
+    dockerfile               = string
+    image_name               = string
+    image_release_tag_prefix = string
+  }))
+  default = [
```
erikschlegel: I feel like this variable should be required and not optional. Things like `app_name` and `image_name` would vary for each customer deployment.

infra/templates/az-isolated-service-single-region/variables.tf
```diff
@@ -38,9 +38,31 @@ variable "ase_vnet_name" {
...
+    {
+      app_name                 = "cobalt-backend-api-1"
+      repository               = "https://github.com/erikschlegel/echo-server.git"
+      dockerfile               = "Dockerfile"
+      image_name               = "appsvcsample/static-site-1"
```
erikschlegel: The source image is an `echo-server` 😄

infra/modules/providers/azure/container-registry/output.tf
iphilpot: Is there a way to export this as a password... ie *************

infra/templates/az-isolated-service-single-region/app.tf
nmiodice: This is described in the comment above but only premium SKUs allow us to configure network ACLs. Any other value for this will result in an error from Azure when attempting to configure the VNET integration.

infra/templates/az-isolated-service-single-region/app.tf
nmiodice: If the initial step fails then the integration tests should catch the error.

infra/templates/az-isolated-service-single-region/variables.tf
nmiodice: I can remove the default value and just supply it during our testing.

infra/templates/az-isolated-service-single-region/app.tf
nmiodice: We don't support other use cases at the moment so if this is not `true` then it will result in a failure. 

Options like this seem useful for modules because they make them flexible, but not as useful for the templates which have a more opinionated stance on the configuration.

infra/templates/az-isolated-service-single-region/ase.tf
nmiodice: I don't think that using a data reference will help us out here as the data reference is also stored in the terraform state.

Here is a trivial example:
```bash
# A test terraform template
$ cat main.tf
terraform {
  backend "azurerm" {
    key = "terraform.tfstate"
  }
}


data "azurerm_container_registry" "acr" {
  name                = "isolatedservicecobaltisolatedtestingacr"
  resource_group_name = "isolated-service-cobalt-isolated-testing-app-rg"
}

# Apply
$ terraform apply
data.azurerm_container_registry.acr: Refreshing state...

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

Now here is a (condensed) version of the state file...
```bash
{
  "resources": [
    {
      "type": "azurerm_container_registry",
      "name": "acr",
      "instances": [
        {
          "attributes": {
            "admin_enabled": true,
            "admin_password": *** CLEAR TEXT PASSWORD REMOVED BY ME ***
            "admin_username": *** CLEAR TEXT USERNAME REMOVED BY ME ***
          }
        }
      ]
    }
  ]
}
```

infra/modules/providers/azure/container-registry/output.tf
nmiodice: This information is stored in the state file either way. You can mark the output as sensitive by using a `sensitive   = true` flag, which I'll add. This only obfuscates the value in CLI output.

infra/templates/az-isolated-service-single-region/app.tf
nmiodice: ACR Build tasks do not seem to work with private repositories [Github Issue](https://github.com/MicrosoftDocs/azure-docs/issues/7975)

However, the source location can be local. Assuming we are able to pull the source down *somehow* we should be able to specify this with the last argument:

```
<SOURCE_LOCATION>
The local source code directory path (e.g., './src') or the URL to a git repository (e.g., 'https://github.com/Azure-Samples/acr-build-helloworld-node.git') or a remote tarball (e.g., 'http://server/context.tar.gz').
```

Noob question: I thought a PAT was only needed to push *into* a repo. Whether or not we can pull into the repo is based on the repos visibility (public|private). Is this correct?

infra/modules/providers/azure/container-registry/output.tf
erikschlegel: When I made my earlier comment, I was unaware that exported TF data source attributes are saved in the state files. Even more shocked that the password attribute for the ACR data source provider isn't marked as sensitive. Your approach makes sense. Thanks for being so thorough with validating the behavior for both approaches.

Issue #201, opened 2019-07-01
Implement AuthN via App Service Module


TechnicallyWilliams:
### Description
As a developer, I'd like to enable EasyAuth on my App Service, so that I can authenticate users.

### Acceptance Criteria
- [x] Should configure EasyAuth using `auth_settings` code block in TF docs in existing App Service Module
- [x] Should use AAD Application ID in `auth_settings` code block 

Also, here are a few points that need to be addressed:

1. Current App Service Module will need to be changed to support `auth_settings`
1. We're only focusing on Web Applications not native applications for the AAD Application type.
1. This user story does not include creating AAD Application through Service Principal module. Instead, it includes using an existing AAD Application from a different tenant.
1. App Service Authentication registration should key off of enabling multi-tenant.

### Resources
[Technical Design Doc](https://github.com/microsoft/cobalt/blob/master/design-reference/architecture/authn/providers/azuread/easyauth.md)
[Terraform App Service Auth Docs](https://www.terraform.io/docs/providers/azurerm/r/app_service.html#auth_settings)
[Terraform AzureAD App Docs](https://www.terraform.io/docs/providers/azuread/r/application.html)

#### App Service Module
1. Create `auth_settings` code block that contains `active_directory` code block.
2. Test deployment with an existing AAD application ID from a different tenant. Templatize the tenant GUID and client id.
3. Update all references to app-service-name in main project due to type change from map to nested map.
4. Update all references to app-service-name in testing suite due to type change from map to nested map.

#### Integration Test
1. Test by using a "throw away" Template that creates an AAD Application and then references this to configure Easy Auth in App Service. Verify in the portal.
    1. Doesn't need to be in an existing template... Just create a template to test and take a screenshot.

erikschlegel:
@TechnicallyWilliams please add the t-shirt sized estimation label for this user story.

PR #211, opened 2019-07-15
Provision app service resource with multi-tenant azure ad authentication

TechnicallyWilliams:
## All Submissions:

Current behavior deploys App Service resources without the choice to opt into authentication. This results in an App Service application that doesn't prompt users to provide credentials in order to gain access.

Issue Number: #201

## What is the new behavior?

- App Service provides Cobalt users the option to opt-in to Azure AD managed authentication (easyauth).
- To opt-in to authentication (easyauth), App Service expects users to supply data from an already preregistered Azure AD application.
- When opting into authentication, Cobalt users have the choice of choosing an already registered Azure AD application that resides in the deploying tenant or an external tenant.

- [NO]

Secondary tenant being used to manage authentication for app service living in another tenant is circled in the below screenshot:
image

## Other information

WIP PR. Still need to rebase.


infra/modules/providers/azure/app-service/main.tf
```diff
@@ -1,7 +1,10 @@
+data "azurerm_client_config" "current" {}
+
 locals {
   access_restriction_description = "blocking public traffic to app service"
   access_restriction_name        = "vnet_restriction"
   acr_webhook_name               = "cdwebhook"
+  tenant_id                      = var.external_tenant_id == "" ? data.azurerm_client_config.current.tenant_id : var.external_tenant_id
```
nmiodice: Recommend that you add locals that looks something like this:
```hcl
app_names  = keys(var.app_service_config)
app_configs = values(var.app_service_config)
```

I think this will improve readability throughout the module. I'll call out areas where it may help in the PR

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -14,25 +17,34 @@ data "azurerm_app_service_plan" "appsvc" {
 }
 
 resource "azurerm_app_service" "appsvc" {
-  name                = "${lower(element(keys(var.app_service_name), count.index))}-${lower(terraform.workspace)}"
+  name                = "${lower(element(keys(var.app_service_config), count.index))}-${lower(terraform.workspace)}"
```
nmiodice: if taking the above suggestion this would look like:
`format("%s-%s", lower(locals.app_names[count.index]), lower(terraform.workspace))`


Also the use of `element()` is unnecessary in terraform 12 I think. This could become:
`"${lower(keys(var.app_service_config)[count.index])}-${lower(terraform.workspace)}"`

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -14,25 +17,34 @@ data "azurerm_app_service_plan" "appsvc" {
...
   resource_group_name = data.azurerm_resource_group.appsvc.name
   location            = data.azurerm_resource_group.appsvc.location
   app_service_plan_id = data.azurerm_app_service_plan.appsvc.id
   tags                = var.resource_tags
-  count               = length(keys(var.app_service_name))
+  count               = length(keys(var.app_service_config))
```
nmiodice: if taking the above suggestion this would look like:
`length(locals.app_names)`

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -14,25 +17,34 @@ data "azurerm_app_service_plan" "appsvc" {
...
     WEBSITES_ENABLE_APP_SERVICE_STORAGE = false
     DOCKER_REGISTRY_SERVER_USERNAME     = var.docker_registry_server_username
     DOCKER_REGISTRY_SERVER_PASSWORD     = var.docker_registry_server_password
     APPINSIGHTS_INSTRUMENTATIONKEY      = var.app_insights_instrumentation_key
     KEYVAULT_URI                        = var.vault_uri
     DOCKER_ENABLE_CI                    = var.docker_enable_ci
+  }        
+
+  auth_settings {
+    enabled            = var.enable_auth
```
nmiodice: Could this be inferred off of the value of `ad_client_id` such that we can reduce the # of variables for the template to a minimum? It may look something like this:
`locals.app_configs[count.index].ad_client_id == "" ? false : true`

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -14,25 +17,34 @@ data "azurerm_app_service_plan" "appsvc" {
...
+    issuer             = format("https://sts.windows.net/%s", local.tenant_id)
+    default_provider   = "AzureActiveDirectory"
+    active_directory {
+      client_id = element(values(var.app_service_config), count.index).ad_client_id
```
nmiodice: `values(var.app_service_config)[count.index].ad_client_id`
or
`locals.app_configs[count.index].ad_client_id`


I'll stop making comments about usage of locals and/or removing `element()` but there are a few more instances where changing this up will help improve the readability of the codebase in general (note: we should have done this a while ago!)

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -14,25 +17,34 @@ data "azurerm_app_service_plan" "appsvc" {
...
+    }
   }
 
   site_config {
-    linux_fx_version     = "DOCKER|${var.docker_registry_server_url}/${element(values(var.app_service_name), count.index)}"
+    linux_fx_version     = format("DOCKER|%s/%s", var.docker_registry_server_url, element(values(var.app_service_config), count.index).image)
```
nmiodice:
```suggestion
    linux_fx_version     = format("DOCKER|%s/%s", var.docker_registry_server_url, locals.app_configs[count.index].image)
```

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -43,23 +55,23 @@ resource "azurerm_app_service" "appsvc" {
 }
 
 resource "null_resource" "acr_webhook_creation" {
-  count      = var.docker_enable_ci == true && var.azure_container_registry_name != "" ? length(keys(var.app_service_name)) : 0
+  count      = var.docker_enable_ci == true && var.azure_container_registry_name != "" ? length(keys(var.app_service_config)) : 0
   depends_on = [azurerm_app_service.appsvc]
 
   triggers = {
-    images_to_deploy = "${join(",", values(var.app_service_name))}"
+    images_to_deploy = "${join(",", [for k, v in var.app_service_config : v.image])}"
```
nmiodice: Really cool! haven't seen this before

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -43,23 +55,23 @@ resource "azurerm_app_service" "appsvc" {
...
     acr_name         = var.azure_container_registry_name
   }
 
   provisioner "local-exec" {
-    command = "az acr webhook create --registry ${var.azure_container_registry_name} --name ${replace(lower(element(keys(var.app_service_name), count.index)), "-", "")}${replace(lower(terraform.workspace), "-", "")}${local.acr_webhook_name} --actions push --uri $(az webapp deployment container show-cd-url -n ${lower(element(keys(var.app_service_name), count.index))}-${lower(terraform.workspace)} -g ${data.azurerm_resource_group.appsvc.name} --query CI_CD_URL -o tsv)"
+    command = format("az acr webhook create --registry %s --name %s%s%s --actions push --uri $(az webapp deployment container show-cd-url -n %s-%s -g %s --query CI_CD_URL -o tsv)", var.azure_container_registry_name, replace(lower(element(keys(var.app_service_config), count.index)), "-", ""), replace(lower(terraform.workspace), "-", ""), local.acr_webhook_name, lower(element(keys(var.app_service_config), count.index)), lower(terraform.workspace), data.azurerm_resource_group.appsvc.name)
```
nmiodice: Can this be broken onto multiple lines to improve readability? Unsure if terraform allows this.

Also consider this approach as I found it to improve readability by adding names to the string substitutions:
https://www.terraform.io/docs/provisioners/local-exec.html#environment
```hcl
resource "aws_instance" "web" {
  # ...

  provisioner "local-exec" {
    command = "echo $FOO $BAR $BAZ >> env_vars.txt"

    environment = {
      FOO = "bar"
      BAR = 1
      BAZ = "true"
    }
  }
}
```

infra/modules/providers/azure/app-service/variables.tf
```diff
@@ -20,12 +20,24 @@ variable "resource_tags" {
   default     = {}
 }
 
-variable "app_service_name" {
+variable "app_service_config" {
   description = "The name of the app service to be created"
-  type        = map(string)
+  type        = map(map(string))
```
nmiodice: We should enforce the schema of this map by leveraging a different format for the type. e.g. https://github.com/microsoft/cobalt/blob/master/infra/templates/az-isolated-service-single-region/variables.tf#L41

This allows terraform to perform type checking that fails invalid configurations at build time rather than runtime

infra/modules/providers/azure/app-service/variables.tf
```diff
@@ -20,12 +20,24 @@ variable "resource_tags" {
...
   default     = {}
 }
 
+variable "enable_auth" {
```
nmiodice: See note about removing this in favor of defaulting to the app-specific config.

infra/templates/az-hello-world/outputs.tf
```diff
@@ -1,4 +1,4 @@
 output "app_service_default_hostname" {
-  value = "http://${element(module.app_service.app_service_uri, 0)}"
+  value = "https://${element(module.app_service.app_service_uri, 0)}"
```
nmiodice: Does this template use HTTPS?

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: Brilliant Jerry! Brilliant!

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: Your suggestion makes way more sense because it scopes authorization to the app level and away from being a master switch that enables authentication for all. Thanks!!

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: With fresh eyes, i'm actually surprised this worked. I'm definitely going to have to look at this again. Thanks for noticing though.

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: Yeah, I'm so glad you're pointing this out too. Yeah, I'm going to look for a better alternative. Thanks for giving resources to start with!

infra/modules/providers/azure/app-service/variables.tf
TechnicallyWilliams: Another great call out! Learning a lot! Thanks!

infra/templates/az-hello-world/outputs.tf
TechnicallyWilliams: For some reason, app service output url in the portal is default https. I investigated, ran a few tests and couldn't figure out why but I made the code match what you would see if visiting your app service "overview" panel in the portal.

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: Making this change was interesting because I ran into errors trying to properly place the string functions (i.e. replace(lower(local.app_names[count.index]), "-", "")).  I opted to place all the sting functions within the environment block. I think this is a situation where one should one place for all string functions. Choose either the 'command' or the 'environment' block but not both because that's prone to errors.

```hcl
  # ...
 provisioner "local-exec" {
    command = "az acr webhook create --registry $ACRNAME --name $APPNAME$WRKSPACE$WEBHOOKNAME . . ."
    
    environment = {
      ACRNAME      = var.azure_container_registry_name
      APPNAME      = replace(lower(local.app_names[count.index]), "-", "")
   }
  }
```

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: Due to splitting up the map, this is now an array that holds configurations without the keys.

```hcl
images_to_deploy = "${join(",", [for config in local.app_configs : config.image])}"
```

infra/modules/providers/azure/app-service/variables.tf
TechnicallyWilliams: app-specific config implemented. Enable_auth removed.

infra/modules/providers/azure/app-service/variables.tf
TechnicallyWilliams: Replaced with the following type:
```hcl
variable "app_service_config" {
  description = "The name of the app service to be created"
  type = map(object({
    image        = string
    ad_client_id = string
  }))
  default     = {}
}
```

infra/templates/az-isolated-service-single-region/ase.tf
```diff
@@ -49,9 +49,13 @@ module "app_service" {
...
   docker_registry_server_username  = module.container_registry.admin_username
   docker_registry_server_password  = module.container_registry.admin_password
-  app_service_name = {
+  external_tenant_id               = var.external_tenant_id
+  app_service_config = {
     for target in var.deployment_targets :
-    target.app_name => "${target.image_name}:${target.image_release_tag_prefix}-${lower(terraform.workspace)}"
+    target.app_name => {
+      image        = "${target.image_name}:${target.image_release_tag_prefix}-${lower(terraform.workspace)}"
+      ad_client_id = "${target.auth_client_id}"
```
nmiodice: nit: unnecessary interpolation syntax

TechnicallyWilliams: Removed for ad_client_id

PR #230, opened 2019-08-04
updated az isolated ase template with create authn feature

TechnicallyWilliams:
## All Submissions:
* [NO] All new and existing tests passed.

Issue Number: #189 

Current behavior would not allow an app service to opt-in to authentication using an az AD auth registration app created from within the same deployment. The az AD auth reg app needed to exist 
 prior to deployment. 

## What is the new behavior?

- az ase template will now create an azure AD auth registration app for each deployment target  from an auth target list. Using an pre-existing azure AD auth app is no longer an option.

- [No]

## Other information

Screenshot of authn working with latest module implementation.
image

### DIFF
--- a/infra/modules/providers/azure/ad-application/output.tf
+++ b/infra/modules/providers/azure/ad-application/output.tf
@@ -1,14 +1,9 @@
-output "application_id" {
-  description = "The client id output by the app registration."
-  value       = azuread_application.auth.*.application_id
-}
-
 output "azuread_config_data" {
-  description = "Output data that pairs azuread names with their client ids."
+  description = "Output data that pairs azuread names with their application ids."
   value = {
     for azuread in data.azuread_application.auth :
     azuread.name => {
-      client_id = azuread.application_id
+      application_id = azuread.application_id
     }
   }
 }
--- a/infra/modules/providers/azure/ad-application/output.tf
+++ b/infra/modules/providers/azure/ad-application/output.tf
@@ -1,0 +1,13 @@
+output "application_id" {
+  description = "The client id output by the app registration."
+  value       = azuread_application.auth.*.application_id
+}
+
+output "azuread_config_data" {
+  value = {
+    for azuread in data.azuread_application.auth :
+    azuread.name => {
+      client_id      = azuread.application_id
+    }
+  }
+}
--- a/infra/modules/providers/azure/ad-application/output.tf
+++ b/infra/modules/providers/azure/ad-application/output.tf
@@ -4,10 +4,11 @@ output "application_id" {
 }
 
 output "azuread_config_data" {
+  description = "Output data that pairs azuread names with their client ids."
   value = {
     for azuread in data.azuread_application.auth :
     azuread.name => {
-      client_id      = azuread.application_id
+      client_id = azuread.application_id
     }
   }
 }
--- a/infra/modules/providers/azure/app-service/main.tf
+++ b/infra/modules/providers/azure/app-service/main.tf
@@ -92,6 +92,12 @@ resource "azurerm_app_service_slot" "appsvc_staging_slot" {
   depends_on          = [azurerm_app_service.appsvc]
 }
 
+data "azurerm_app_service" "all" {
+  count               = length(azurerm_app_service.appsvc)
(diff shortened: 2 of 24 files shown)


infra/modules/providers/azure/ad-application/output.tf
```diff
@@ -0,0 +1,14 @@
+output "application_id" {
```
awkwardindustries: Because we're outputting a map of azuread_application.auth.name to the azuread_application.auth.application_id, do we really need this also? I prefer the mapped version since we're possibly creating more than one app registration and it provides flexibility of more context.

infra/modules/providers/azure/ad-application/output.tf
```diff
@@ -0,0 +1,14 @@
...
+  description = "The client id output by the app registration."
+  value       = azuread_application.auth.*.application_id
+}
+
+output "azuread_config_data" {
+  description = "Output data that pairs azuread names with their client ids."
+  value = {
+    for azuread in data.azuread_application.auth :
+    azuread.name => {
+      client_id = azuread.application_id
```
awkwardindustries: Would recommend leaving at application_id (instead of client_id =, use application_id =) since the client_id is from the perspective of the consumer of this config data.

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -37,18 +37,19 @@ resource "azurerm_app_service" "appsvc" {
   }
 
   auth_settings {
-    enabled          = local.app_configs[count.index].ad_client_id == "" ? false : true
+    enabled          = false
```
awkwardindustries: Since we're always disabling the auth_settings, are we able to eliminate this block altogether? It's counter intuitive that we're half setting up for the possibility that it *might* be used for authn, but not necessarily required.

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -37,18 +37,19 @@ resource "azurerm_app_service" "appsvc" {
...
-      client_id = local.app_configs[count.index].ad_client_id
+      client_id = ""
     }
   }
 
   site_config {
     linux_fx_version     = format("DOCKER|%s/%s", var.docker_registry_server_url, local.app_configs[count.index].image)
     always_on            = var.site_config_always_on
     virtual_network_name = var.vnet_name
+    scm_type             = "None"
```
awkwardindustries: What's the intent behind adding this change to the site_config?

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -91,6 +92,12 @@ resource "azurerm_app_service_slot" "appsvc_staging_slot" {
   depends_on          = [azurerm_app_service.appsvc]
 }
 
+data "azurerm_app_service" "all" {
```
awkwardindustries: Is this required? I see that the azurerm_template_deployment.access_restriction depends on this, but it doesn't look as though the data is actually used otherwise.

TechnicallyWilliams: I think I gave the impression that this wasn't needed but it's actually used in the outputs.tf of the iso template.

infra/modules/providers/azure/ad-application/output.tf
TechnicallyWilliams: This was removed.

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: I thought not having this set to "none" caused some local issues I was having that displayed username/password information in the terraform plan. Turns out, its not an issue in the pipeline. Since this defaults to "none", I removed this line after all.

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: Removing this and adding the tenant id from the azure cli worked out. Below is a screenshot of visiting a basic web app with authentication.

image

infra/modules/providers/azure/ad-application/output.tf
TechnicallyWilliams: client_id was replaced with application_id in all places.

Issue #252, opened 2019-08-26
Create new version and go process


iphilpot:
### Spike Description
The goal of the spike is to perform and document operations that support use cases, listed below, which target three personas: DevOps Engineer, App Team Developer, and App Team Release Manager. In general, the goal is to use Cobalt without having to use the 'fork & go' method of adoption, but rather create and support an AZ DO Pipeline configuration that points directly to the App Team repo (vs 'Importing' the `microsoft/cobalt` repo, or forks of that repo).

As a user, I'd like to source a template to a version and use this in my code for deployments, in order to have a simple deployment process

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)

- [ ] Should use version for source of template
- [ ] Should have a simple way to deploy using existing CI/CD
- [ ] Should have configurable variables provided to app dev team

#### Support the following use cases:
- [ ] DevOps-Eng: onboard an app-team into a Cobalt new AZDO pipeline project
    * DevOps-Eng sends App-Dev the `infra` folder
    * App-Dev puts `infra` at the root of their git repo, adjusts values therein, 
    pushes change to `master`. Then notifies DevOps.
    * DevOps-Eng creates new AZDO Project, points to `infra`, does a `devint` deployment, 
    perhaps `qa` and/or `release` envs, too.
    * DevOps-Eng notifies App-Team to use the new pipeline for CI/CD, releases, etc.
- [ ] App-Dev commits & PRs code, merges to `master`, wants new WIP/RC build to go live on `devint` or perhaps `qa`, etc.
- [ ] App-Dev uses the AZ DO pipeline to trigger a deployment to the desired env
- [ ] App-Release-Manager wants to release a new version of App to production
    * AppReleaseManager uses the AS DO Pipeline to trigger a production deployment
- [ ] DevOps Guy: wants to get App-Team to use a new version of Cobalt and/or it's templates
    * DevOp-Eng, or App-Dev, alters the `infra` files to point to a new tagged release of either the 
    Cobalt bits (either at `microsoft/cobalt@vNext//infra/templates/az-service-single-region` or better, 
    at `foo.org@v6//cobalt-template-az-service-single-region` 
    or maybe `foo.org@v6//co-template-std-microservice`) *

Also, here are a few points that need to be addressed:

1. Keep the scope of work tight (it is a spike). Throw-away code. Keep docs (place in ``).
[...]

PR #260, opened 2019-09-02
Adds a design document for the version and go process

iphilpot:
## All Submissions:
* [x] Have you followed the guidelines in our Contributing [document](./CONTRIBUTING.md)?
* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [x] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] I ran lint checks locally prior to submission.
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Currently, we are using a fork and go process.

Issue Number: #252 

## What is the new behavior?
No new current behavior design documentation for Version and Go process. This is intended to replace Fork-n-go.

- [No]

n/a

## Other information
n/a

### DIFF
--- a/design-reference/architecture/VersionAndGo/readme.md
+++ b/design-reference/architecture/VersionAndGo/readme.md
@@ -3,15 +3,15 @@
 This design addresses Cobalt's workflow to deploy individual application infrastructure using a Cobalt template.
 
 ## Definitions
- - `Engineering Team (Eng Team)` - DevOps or Operations team consisting of Infrastructure Administrators or similar roles.
+ - `Operations Team (Ops Team)` - DevOps or Operations team consisting of Infrastructure Administrators or similar roles.
  - `Application Team (App Team)` - Software Developers with an application to deploy to Azure.
  - `Development Repo (Dev Repo)` - Cobalt's repo hosted on Github at [https://github.com/microsoft/cobal](https://github.com/microsoft/cobal)
- - `Engineering Repo (Eng Repo)` - Individual template repository managed by the `Engineering Team`
+ - `Operations Repo (Ops Repo)` - Individual template repository managed by the `Operations Team`
  - `Application Repo (App Repo)` - Repository that contains the application code to be deployed and managed by the `Application Team`
- - `Version` - all versioning schemes should use [SemVer](https://semver.org/) or variant. 
+ - `Version` - all versioning schemes should use [SemVer](https://semver.org/) or variant. [ref: Terraform Versioning](https://www.terraform.io/docs/extend/best-practices/versioning.html)
 
 ## Version and Go Setup Description
-Our design starts after the initial Cobalt setup, where an individual template is cloned into an organizations source control (ref: [Setup Doc](https://github.com/microsoft/cobalt/blob/master/docs/GETTING_STARTED_APP_DEV.md)). An application development team is ready to start deploying their code to infrastructure built in Azure. Typically they would make a request to the engineering team. The Eng team would provide the App team with a link to the latest infra configuration files for a specific template. The App team downloads this tar/zip file and extracts the folder/files.
+Our design starts after the initial Cobalt setup, where an individual template is cloned into an organizations source control (ref: [Setup Doc](https://github.com/microsoft/cobalt/blob/master/docs/GETTING_STARTED_APP_DEV.md)). An application development team is ready to start deploying their code to infrastructure built in Azure. Typically they would make a request to the Operations Team. The Ops Team would provide the App team with a link to the latest infra configuration files for a specific template. The App team downloads this tar/zip file and extracts the folder/files.
 
 ```
 infra/
@@ -19,7 +19,7 @@ infra/
  |_ Application.yaml
 ```
 
-These files should be derived from the template folder in dev repo and cloned into the eng repo. Proposed folder path/location is:
+These files could be derived from the template folder in dev repo and cloned into the Ops Repo. Proposed folder path/location is:
 
 ```
 infra/
@@ -28,13 +28,23 @@ infra/
           |_tests
           |_config
           |   |_ application.tf
-          |   |_ application.yml
           |_template files...
 ```
 
- - `Appliation.tf` is the artifact that points to the template in the eng repo. The application team would need to fill this file out with the requested configuration. Example:
+```
+devops/
+  |_ providers
+      |_ azure-devops
+          |_ templates
+              |_ infrastructure
+                  |_ application.yaml
+```
 
- ```
+We could also approach this using documentation. The idea would be it's the organizations responsibility to create or provide guidance on creating. This would simplify the overall design.
+
+ - `Appliation.tf` is the artifact that points to the template in the Ops Repo. The application team would need to fill this file out with the requested configuration. Example:
+
+ ```hdl
  module "az-isolated-service-single-region" {
    source = "git://dev.azure.com/project/repo?ver=1.0.0//infra/template/az-isolated-service-single-region/"
    resource_group_location = "eastus"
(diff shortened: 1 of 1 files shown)


design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
+# Version and Go Technical Design Purpose
+
+This design addresses Cobalt's workflow to deploy individual application infrastructure using a Cobalt template.
+
+## Definitions
```
nmiodice: Really like that you included this section as it makes the document easier to read, especially after not looking at it for a while.

One minor nit is that I think the breakdown of `engineering` vs `application` teams is a bit difficult to grok mentally since I could see either team calling themselves an `engineering` team.

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+  |_template
+      |_template_name
+          |_tests
+          |_config
+          |   |_ application.tf
+          |   |_ application.yml
+          |_template files...
+```
+
+ - `Appliation.tf` is the artifact that points to the template in the eng repo. The application team would need to fill this file out with the requested configuration. Example:
```
nmiodice: I really like this approach as it minimizes the content that the application teams need to pull in as dependencies. This also makes it easier to understand which versions teams in the case that there are migration efforts to move to the latest infrastructure best practices within an organization.

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+This design addresses Cobalt's workflow to deploy individual application infrastructure using a Cobalt template.
+
+## Definitions
+ - `Engineering Team (Eng Team)` - DevOps or Operations team consisting of Infrastructure Administrators or similar roles.
+ - `Application Team (App Team)` - Software Developers with an application to deploy to Azure.
+ - `Development Repo (Dev Repo)` - Cobalt's repo hosted on Github at [https://github.com/microsoft/cobal](https://github.com/microsoft/cobal)
+ - `Engineering Repo (Eng Repo)` - Individual template repository managed by the `Engineering Team`
+ - `Application Repo (App Repo)` - Repository that contains the application code to be deployed and managed by the `Application Team`
+ - `Version` - all versioning schemes should use [SemVer](https://semver.org/) or variant.
```
nmiodice: We may want to refer to this documentation that is published by Terraform: https://www.terraform.io/docs/extend/best-practices/versioning.html

design-reference/architecture/VersionAndGo/readme.md
nmiodice: I don't think that we have a dependency on this living in a repo specifically. If we indeed don't need to version via git we could refer to this documentation and leave it up to the organization to choose what's best for them: https://www.terraform.io/docs/modules/sources.html

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+ - `Application.yaml` - is the artifact that points to the azure-pipeline.yml file in the eng repo that contains the build/test/deploy pipeline.
+
+This folder/files should be checked into the root of the app repo. The Eng team would then configure a new pipeline in the app teams Azure Devops project to point to these files and deploy the requested infrastructure.
+
+> Note: We should strive to continue to use a single set of pipeline files for all pipelines referenced in this document.
+
+## Version and Go Eng Pipeline Description
+
+This pipeline will produce an artifact that is stored in a location that would be accessible by both the engineering team and the app team. It should follow or use the same pipeline as the development team and app team pipelines. This code that publishes a new version could be controlled by a feature flag as global variable in the eng pipeline. This could also use a SemVer global variable and only create a new artifact when this has been incremented. Would could also leave this as a "manual" process. Manual only means that the pipeline doesn't create the artifact. We could use a script or the likes to do this.
```
nmiodice: Note: not entierly sure I understand this section, so my questions may be off-base a bit. 

> This pipeline will produce an artifact that is stored in a location that would be accessible by both the engineering team and the app team

Are you saying that this pipeline essentially produces a new "version" (artifact) of the terraform source module that can be consumed by both teams? I.e., publish to terraform registry, create a tag in git, etc...?

> It should follow or use the same pipeline as the development team and app team pipelines

Can you elaborate on why? The infrastructure focused pipelines follow a pattern of terraform commands & tests that execute in a given environment. The pipeline described here seems to be a version publish. Unclear on how much re-use there is between these steps - can you help me understand?

> controlled by a feature flag

Will this amount to the following? If so, I'm not seeing the benefit of using a single pipeline definition.

```
if shouldPublishNewVersion == True:
    < run publish workflow >
else:
    < deploy to $STAGE >
```

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+}
+ ```
+
+ - `Application.yaml` - is the artifact that points to the azure-pipeline.yml file in the eng repo that contains the build/test/deploy pipeline.
+
+This folder/files should be checked into the root of the app repo. The Eng team would then configure a new pipeline in the app teams Azure Devops project to point to these files and deploy the requested infrastructure.
+
+> Note: We should strive to continue to use a single set of pipeline files for all pipelines referenced in this document.
+
+## Version and Go Eng Pipeline Description
```
nmiodice: Not sure this is in scope for our milestone but I could see these be different yaml templates for the different publishing strategies. i.e., one for "publish to git", one to "publish to terraform repo", etc...

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+This pipeline will produce an artifact that is stored in a location that would be accessible by both the engineering team and the app team. It should follow or use the same pipeline as the development team and app team pipelines. This code that publishes a new version could be controlled by a feature flag as global variable in the eng pipeline. This could also use a SemVer global variable and only create a new artifact when this has been incremented. Would could also leave this as a "manual" process. Manual only means that the pipeline doesn't create the artifact. We could use a script or the likes to do this.
+
+## Version and Go App Pipeline Description
+
+The pipeline "bootstrap" file: `Application.yaml` should perform the following...
+
+1. Request to copy the pipeline files down to the build agent from the engineering repo.
```
nmiodice: is this a source code checkout? Unclear what "pipeline files" are. Scripts used by the templates? If so, i'm interested to learn how AzDO handles this

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+2. It should use the local `Application.tf` file to "bootstrap" the template and download all needed template/module files.
+3. Deploy the template it that is referenced by the `Application.tf` file using the pipeline referenced by the `Application.yml` file.
+
+# Risks and Unknowns
+
+1. Local relative path used in the template to source the modules may fail if we do not copy the module files to the build agent.
+2. We need to add specific functionality to the engineering pipeline to publish versioned artifacts.
```
nmiodice: IMO this is probably worth automating further in the future than the rest of what is described in this document, and I think it should live in a separate pipeline definition as well.

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+3. We need to be able to copy the same version of pipeline files as the template.
+4. We need to configure the pipeline to use `Application.tf` not template `main.tf ` to deploy from the app repo.
```
nmiodice: Terraform doesn't care about this -- we should be good to go w/o any work.

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+These files should be derived from the template folder in dev repo and cloned into the eng repo. Proposed folder path/location is:
+
+```
+infra/
+  |_template
+      |_template_name
+          |_tests
+          |_config
+          |   |_ application.tf
+          |   |_ application.yml
```
nmiodice: If our goal is to use the same template file across any infrastructure deployment, might it make sense to move this to a directory that is not template specific?

Note: you can also reference templates from another repository so we may not even need app dev teams to clone this file into their repository because it would be a trivial template import: 
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#using-other-repositories

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+These files should be derived from the template folder in dev repo and cloned into the eng repo. Proposed folder path/location is:
+
+```
+infra/
+  |_template
+      |_template_name
+          |_tests
+          |_config
+          |   |_ application.tf
```
nmiodice: Could this be replaced with good documentation on how to use the template?

If yes, and if we move the template to another location (see below comment) then we may not need to have any handoff for the client teams, removing the "tar/zip" step completely.

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+ - `Application Team (App Team)` - Software Developers with an application to deploy to Azure.
+ - `Development Repo (Dev Repo)` - Cobalt's repo hosted on Github at [https://github.com/microsoft/cobal](https://github.com/microsoft/cobal)
+ - `Engineering Repo (Eng Repo)` - Individual template repository managed by the `Engineering Team`
+ - `Application Repo (App Repo)` - Repository that contains the application code to be deployed and managed by the `Application Team`
+ - `Version` - all versioning schemes should use [SemVer](https://semver.org/) or variant. 
+
+## Version and Go Setup Description
+Our design starts after the initial Cobalt setup, where an individual template is cloned into an organizations source control (ref: [Setup Doc](https://github.com/microsoft/cobalt/blob/master/docs/GETTING_STARTED_APP_DEV.md)). An application development team is ready to start deploying their code to infrastructure built in Azure. Typically they would make a request to the engineering team. The Eng team would provide the App team with a link to the latest infra configuration files for a specific template. The App team downloads this tar/zip file and extracts the folder/files.
```
nmiodice: > The Eng team would provide the App team with a link to the latest infra configuration files for a specific template. The App team downloads this tar/zip file and extracts the folder/files.

See below for ideas about how to avoid this step, which would lower the burden & manual effort for the engineering teams

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+
+This pipeline will produce an artifact that is stored in a location that would be accessible by both the engineering team and the app team. It should follow or use the same pipeline as the development team and app team pipelines. This code that publishes a new version could be controlled by a feature flag as global variable in the eng pipeline. This could also use a SemVer global variable and only create a new artifact when this has been incremented. Would could also leave this as a "manual" process. Manual only means that the pipeline doesn't create the artifact. We could use a script or the likes to do this.
+
+## Version and Go App Pipeline Description
+
+The pipeline "bootstrap" file: `Application.yaml` should perform the following...
+
+1. Request to copy the pipeline files down to the build agent from the engineering repo.
+2. It should use the local `Application.tf` file to "bootstrap" the template and download all needed template/module files.
+3. Deploy the template it that is referenced by the `Application.tf` file using the pipeline referenced by the `Application.yml` file.
```
nmiodice: > the pipeline referenced by the `Application.yml` file.

A bit confusing since `Application.yml` is the pipeline, and does not reference itself. Perhaps I'm mis-understanding.

design-reference/architecture/VersionAndGo/readme.md
```diff
@@ -0,0 +1,113 @@
...
+5. We need to understand how to version using Azure Repos. 
+6. We need to verify we can link to the version "tag" of the Azure Repo from the `source` property in the `Application.tf` file.
+
+# Out-of-Scope
+
+We are currently not including the following items.
+
+  1. Containerization of the commands executed by the pipeline.
+
+# Sequence Diagram
```
iphilpot: Need to add information on how to update engineering repo and version. How does the app team update?

design-reference/architecture/VersionAndGo/readme.md
iphilpot: The idea is we use a "copy" task to download azure-pipelines.yaml and dependant files from the Operations repo down to the build agent and execute them similar to how we use the build and release templates now.

design-reference/architecture/VersionAndGo/readme.md
iphilpot: I think the fact we can ref pipeline in another repo solves a lot of this.

Issue #264, opened 2019-09-03
Build Version-n-go artifacts


iphilpot:
### Description
As a DevOps Engineer, I'd like to have all the artifacts for version-n-go, in order to easily deploy applications using the engineering repo.

### Acceptance Criteria
- [x] Should contain application.yaml file to configure CI/CD pipeline
- [x] Should contain application.tf file to deploy specific template
- [ ] All three templates deployable using this method

Also, here are a few points that need to be addressed:

1. Azure Hello World is the template for Development 

### Resources
[Technical Design Doc](https://github.com/microsoft/cobalt/blob/bf77a98549523f7adac5602e30af193c7e35c0db/design-reference/architecture/VersionAndGo/readme.md)

### Tasks

- [x] Verify terraforms version-n-go feasibility
- [x] Verify AzDo version-n-go feasibility
- [x] Resolve Git URL issues for GitHub and Azure DevOps Repos
- [x] Research strategy for handling Variable Groups across projects
- [x] Research and implement git credentials strategy
- [x] Create application.tf and test
- [x] Create application.yaml and test
- [ ] Create script to clone app pipeline DevOps objects
- [x] Determine how to share example application.tf / application.yaml files

PR #280, opened 2019-09-05
Update pipelines to support version-n-go process

awkwardindustries:
## All Submissions:
* [x] Have you followed the guidelines in our Contributing [document](./CONTRIBUTING.md)?
* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] I have updated the documentation accordingly.
* [ ] I have added tests to cover my changes.
* [x] All new and existing tests passed.
* [x] My code follows the code style of this project.
* [x] I ran lint checks locally prior to submission.
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Existing Azure DevOps pipelines required tests to always be run, and checked to see if there were changes to files before executing the build/release jobs. It also did not support the version-n-go process because when the build agent attempted to execute a Terraform `init` command, it would fail to pull from private repositories since it did not have valid Git credentials.

Issue Number: #264 #270 

## What is the new behavior?
This PR covers the changes to the pipelines to support the existing behavior in addition to allowing (via environment variable or parameter configuration):
- Optionally skipping unit and integration test execution
  - Environment Variable: `SKIP_TESTS`
  - Pipeline Parameter: `skipTests`
- Optionally forcing the build and release jobs to run
  - Environment Variable: `FORCE_RUN`
  - Pipeline Parameter: `forceRun`
- Optionally setting the build agent's Git configuration to inject a PAT as provided by the `TEMPLATE_REPO_PAT` environment variable
- Addition of an application-specific entry template that handles pulling down expected scripts from the repository and sets up the build/release jobs

No

No

## Other information
Tested existing scenario of setting up the template-specific repository (ref: [Getting Started](https://github.com/microsoft/cobalt/blob/master/docs/GETTING_STARTED_ADD_PAT_OWNER.md)). Also tested against new version-n-go process successfully.

### DIFF
--- a/devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
+++ b/devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
@@ -26,7 +26,7 @@ jobs:
   - job: Build_${{ config.jobName }}_${{ parameters.environment }}
     dependsOn: TemplateChangeDetection
     pool: $(AGENT_POOL)
-    condition: or(coalesce(variables.FORCE_RUN, ${{ parameters.forceRun }}), eq(dependencies.TemplateChangeDetection.outputs['${{ config.jobName }}.needs_cicd'], 'true')
+    condition: or(coalesce(variables.FORCE_RUN, ${{ parameters.forceRun }}), eq(dependencies.TemplateChangeDetection.outputs['${{ config.jobName }}.needs_cicd'], 'true'))
 
     variables:
     - group: '${{ parameters.environment }} Environment Variables'
--- a/devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
+++ b/devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
@@ -1,13 +1,16 @@
 parameters:
-  alwaysRun: 'false'
   environment: ''
   configurationMatrix: []
+  forceRun: false
+  skipTests: false
 
 jobs:
 
 - job: TemplateChangeDetection
+  variables:
+  - group: '${{ parameters.environment }} Environment Variables'
   displayName: Determine CI Targets to Run
-  condition: ne(${{ parameters.alwaysRun }}, 'true')
+  condition: not(coalesce(variables.FORCE_RUN, ${{ parameters.forceRun }}))
   steps:
   - ${{ each config in parameters.configurationMatrix }}:
     - task: Bash@3
@@ -23,7 +26,7 @@ jobs:
   - job: Build_${{ config.jobName }}_${{ parameters.environment }}
     dependsOn: TemplateChangeDetection
     pool: $(AGENT_POOL)
-    condition: or(eq(${{ parameters.alwaysRun }}, 'true'), eq(dependencies.TemplateChangeDetection.outputs['${{ config.jobName }}.needs_cicd'], 'true'))
+    condition: or(coalesce(variables.FORCE_RUN, ${{ parameters.forceRun }}), eq(dependencies.TemplateChangeDetection.outputs['${{ config.jobName }}.needs_cicd'], 'true')
 
     variables:
     - group: '${{ parameters.environment }} Environment Variables'
@@ -66,14 +69,16 @@ jobs:
     - task: Bash@3
       name: SetupGitCredentialInjection
       displayName: Configure Git to use PAT
-      condition: eq(${{ coalesce(config.overrideGitPAT, 'false') }}, 'true')
+      condition: ne(variables.TEMPLATE_REPO_PAT, '')
       inputs:
         targetType: 'inline'
-        script: '[[ -z $(TEMPLATE_REPO_PAT) ]] && echo "No PAT provided; skipping." || git config --global url."https://cobalt:$(TEMPLATE_REPO_PAT)@".insteadOf "https://"'
+        script: '[[ -z $SecretPAT ]] && echo "No PAT provided; skipping." || git config --global url."https://cobalt:$SecretPAT@".insteadOf "https://"'
+      env:
+        SecretPAT: $(TEMPLATE_REPO_PAT)
 
     - task: AzureCLI@1
       displayName: 'Unit Test Terraform Template'
-      condition: ne(${{ coalesce(config.skipTests, 'false') }}, 'true')
+      condition: not(coalesce(variables.SKIP_TESTS, ${{ parameters.skipTests }}))
       inputs:
         azureSubscription: '$(SERVICE_CONNECTION_NAME)'
         scriptPath: './$(BUILD_ARTIFACT_NAME)/$(PIPELINE_ROOT_DIR)/$(SCRIPTS_DIR)/test-unit.sh'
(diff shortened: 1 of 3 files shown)


awkwardindustries:
Note: this PR does not include the `application.yml` and `application.tf` files that would need to be added to the application repository. We can provide them as templates/guides or else document their contents -- whichever approach makes the most sense. Per the design spec, they should be provided to application teams to support the version-n-go process, but this PR does not cover how they are or will be included.

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
```diff
@@ -1,11 +1,13 @@
 parameters:
+  alwaysRun: 'false'
   environment: ''
   configurationMatrix: []
 
 jobs:
 
 - job: TemplateChangeDetection
   displayName: Determine CI Targets to Run
+  condition: ne(${{ parameters.alwaysRun }}, 'true')
```
nmiodice: Can you help me understand why this needed here? Seems like we would always want to run the job and defer to the condition on line 26.

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
```diff
@@ -1,11 +1,13 @@
 parameters:
+  alwaysRun: 'false'
```
nmiodice: A thought, but not necessarially something to change. Does it make sense to (additionally? instead of?) drive this by an env variable so that it can be over-run on a per-deployment basis w/o a code change?

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
```diff
@@ -61,8 +63,17 @@ jobs:
       inputs:
         version: '$(GO_VERSION)'
 
+    - task: Bash@3
+      name: SetupGitCredentialInjection
+      displayName: Configure Git to use PAT
+      condition: eq(${{ coalesce(config.overrideGitPAT, 'false') }}, 'true')
```
nmiodice: I always get confused with boolean logic, but this may be simplified to something like this:
```bash
condition: coalesce(config.overrideGitPAT, 'false')
```

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
```diff
@@ -61,8 +63,17 @@ jobs:
...
+      inputs:
+        targetType: 'inline'
+        script: '[[ -z $(TEMPLATE_REPO_PAT) ]] && echo "No PAT provided; skipping." || git config --global url."https://cobalt:$(TEMPLATE_REPO_PAT)@".insteadOf "https://"'
+
     - task: AzureCLI@1
       displayName: 'Unit Test Terraform Template'
+      condition: ne(${{ coalesce(config.skipTests, 'false') }}, 'true')
```
nmiodice: Similar to above. Also check out the full list of functions if you haven't yet seen them:
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#functions

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
```diff
@@ -83,6 +94,14 @@ jobs:
         TF_VAR_remote_state_account: $(REMOTE_STATE_ACCOUNT)
         TF_VAR_remote_state_container: $(REMOTE_STATE_CONTAINER)
 
+    - task: Bash@3
+      name: TeardownGitCredentialInjection
+      displayName: Reset Git configuration
+      condition: and(always(), eq(${{ coalesce(config.overrideGitPAT, 'false') }}, 'true'))
```
nmiodice: I think that we *always* want to run this if the override for PAT is set, but I don't think the condition matches this. I think you want something like:

```yml
  ${{ if coalesce(config.overrideGitPAT, 'false') }}:
    - task: Bash@3
      name: TeardownGitCredentialInjection
      displayName: Reset Git configuration
      condition: always()
      inputs:
        targetType: 'inline'
        script: '[[ -z $(TEMPLATE_REPO_PAT) ]] && echo "No PAT provided; skipping." || git config --global --unset url."https://cobalt:$(TEMPLATE_REPO_PAT)@".insteadOf'

```

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-release-stage.yml
```diff
@@ -1,11 +1,13 @@
 parameters:
+  alwaysRun: 'false'
   environment: ''
   configurationMatrix: []
 
 jobs:
 
 - job: TemplateChangeDetection
   displayName: Determine CD Targets to Run
+  condition: ne(${{ parameters.alwaysRun }}, 'true')
```
nmiodice: Similar comment as above. Does line 25 take care of this? If always run is false, then our existing change detection won't ever be included in the template

devops/providers/azure-devops/templates/infrastructure/azure-pipelines-app.yml
```diff
@@ -0,0 +1,61 @@
...
+      displayName: Download Cobalt scripts
+      inputs:
+        targetType: 'inline'
+        script: |
+          # Note: This is intentionally inline because applications should only have
+          # a minimal pipeline definition and Terraform module (for Cobalt) in their repo.
+          mkdir cobaltsrcripts
+          cd cobaltsrcripts
+          git init
+          git remote add origin https://github.com/microsoft/cobalt.git
```
nmiodice: Do we want this to point to the client's HLD repo?

devops/providers/azure-devops/templates/infrastructure/azure-pipelines-app.yml
```diff
@@ -0,0 +1,61 @@
...
+          git fetch --force --tags --prune --progress --no-recurse-submodules origin
+          git checkout --progress --force origin/master
```
nmiodice: Do we want to peg this to the tag from which they are versioned to?

devops/providers/azure-devops/templates/infrastructure/azure-pipelines-app.yml
awkwardindustries: Intentionally deferring this for now to be covered by #286 to allow time to account for constraints.

devops/providers/azure-devops/templates/infrastructure/azure-pipelines-app.yml
awkwardindustries: Intentionally deferring this for now to be covered by #286 to allow time to account for constraints.

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
awkwardindustries: Thinking we can save the execution time if we have a variable that will effectively override this check regardless. It's admittedly not much savings, but didn't seem necessary.

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
awkwardindustries: Great suggestion -- will change to an environment variable.

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
awkwardindustries: This will look a bit different once the environment variable changes comes through, but yes -- it could definitely be simplified.

devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
awkwardindustries: As I understand the docs, the *always()* means that it will run even if there's a failure in the pipeline (which is what we would want), but... only if we actually set it. Therefore, the *and* allows us to say if we used a Git credential PAT, always undo that step -- even if there was a pipeline failure, cancellation, or some other issue. If there wasn't a Git credential PAT, you can skip it.

awkwardindustries:
Refactored to handle setting the skipping tests and forcing runs through environment variables, then parameters, then default values (false for both). Allows for modification without requiring a pipeline update while still supporting expected defaults for fork-n-go (false for both) versus version-n-go (true for both) processes.

PR #291, opened 2019-09-10
Altered pipeline jobs to respect AGENT_POOL

ThePhen:
## All Submissions:
* [YES/NO] Have you added an explanation of what your changes do and why you'd like us to include them?
* [YES/NO] I have updated the documentation accordingly.
* [YES/NO/NA] I have added tests to cover my changes.
* [YES/NO/NA] All new and existing tests passed.
* [YES/NO/NA] My code follows the code style of this project.
* [YES/NO/NA] I ran lint checks locally prior to submission.
* [YES/NO] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Several pipeline jobs were not respecting `AGENT_POOL`, thus preventing use of self-hosted build agents.

Issue Number: 288

## What is the new behavior?
Pipeline jobs are now respecting `AGENT_POOL`.

- [NO] however...

It's possible that use of some agent pools will lead to concurrent jobs. It's important that jobs & templates be written in a way that supports concurrency.

n/a

## Other information
n/a

### DIFF
--- a/devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
+++ b/devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
@@ -6,6 +6,7 @@ jobs:
 
 - job: TemplateChangeDetection
   displayName: Determine CI Targets to Run
+  pool: $(AGENT_POOL)
   steps:
   - ${{ each config in parameters.configurationMatrix }}:
     - task: Bash@3
--- a/devops/providers/azure-devops/templates/azure-pipelines.yml
+++ b/devops/providers/azure-devops/templates/azure-pipelines.yml
@@ -38,17 +38,17 @@ stages:
     configurationMatrix:
     - jobName: az_hello_world
       terraformTemplatePath: 'infra/templates/az-hello-world'
-      terraformWorkspacePrefix: '288hw'
+      terraformWorkspacePrefix: 'hw'
       environmentsToTeardownAfterRelease:
       - 'devint'
 
     - jobName: az_service_single_region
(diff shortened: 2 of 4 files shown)


devops/providers/azure-devops/templates/infrastructure/azure-pipeline-build-stage.yml
```diff
@@ -6,6 +6,7 @@ jobs:
 
 - job: TemplateChangeDetection
   displayName: Determine CI Targets to Run
+  pool: $(AGENT_POOL)
```
iphilpot: Adding this, will it cause us to only use custom build agents? Can we still use hosted agents? 

We don't want every customer/template to use custom build agents only, correct?

nmiodice: Follow on -- have we tested with hosted and non hosted agents? I recall there being some nuances in how the `deployment` step is configured with regard to build agents

ThePhen: Great points. I’ll find out. I didn’t try the hosted agent, after this change.

ThePhen: @iphilpot if `AGENT_POOL` is set to `Hosted Ubuntu 1604` then Cobalt will use hosted agents, as before. 

@nmiodice @iphilpot I've tested the hosted agents here (an AzDO Pipeline build):
https://dev.azure.com/sthende/sthende-291/_build/results?buildId=52&view=results

PR #324, opened 2019-09-25
Remove ACR Webhooks; Do not manage container settings via Terraform

nmiodice:
## All Submissions:

ACR Webhooks are determined to be unsuitable for the deployment strategy that we want to use. This is detailed in https://github.com/microsoft/cobalt/issues/316.

This change removes the ACR webhooks and also prevents Terraform from changing the container settings for existing webapps.

Issue Number: https://github.com/microsoft/cobalt/issues/316

## What is the new behavior?
- `az-hello-world` is unchanged
- `az-service-single-region` deploys containerized webapp. The Image is specified after the initial deployment and subsequent terraform deployments will not try to override the container settings. No ACR webhook is configured.
- `az-isolated-service-single-region` deploys containerized webapp. No image is specified after the initial deployment and subsequent terraform deployments will not try to override the container settings. No ACR webhook is configured.

- [YES] ACR Pull no longer works


infra/modules/providers/azure/provider/main.tf
```diff
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version = "~>1.32.0"
+  version = "~>1.34.0"
```
KeithJRome: Is there a changelog available for this? I'm wondering if we need to be on the lookout for any new breaks related to a provider rev.

ThePhen:
Aha! So this was letting the .terraform.d folder hang around between jobs/stages, @nmiodice?

infra/modules/providers/azure/provider/main.tf
ThePhen: Good point. If the upstream is following SemVer, that change should t be breaking (or else the major number would have been incremented).

But, I don’t that they’re using SemVer. I can help check tomorrow, @KeithJRome

nmiodice: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md#1340-september-18-2019

nmiodice: I upgraded specifically because of a bug related to the provider code that was fixed -- https://github.com/terraform-providers/terraform-provider-azurerm/issues/4184

nmiodice:
@ThePhen That was a "try this to see if it works" approach that I ended up backing out. We actually want to keep the `.terraform` folder around as it includes some helpful things. See some more commentary [here](https://learn.hashicorp.com/terraform/development/running-terraform-in-automation#plan-and-apply-on-different-machines) if you are interested

infra/modules/providers/azure/provider/main.tf
KeithJRome: Note that this bumps the Terraform dependency to 12.8 also (https://github.com/terraform-providers/terraform-provider-azurerm/pull/4341)

These changes rolled up into 1.34 look like they are good for us on balance, but I see several behavioral changes that we probably need to be aware of like how certain errors are being reported and some workarounds we have in place now may no longer be needed.

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -32,40 +42,25 @@ resource "azurerm_app_service" "appsvc" {
...
-      WEBHOOKNAME = local.acr_webhook_name
-      APPSVCNAME  = data.azurerm_resource_group.appsvc.name
-    }
 
+  lifecycle {
+    # This stanza will prevent terraform from reverting changes to the application container settings.
+    # These settings are how application teams deploy new containers to the app service and should not
+    # be overridden by Terraform deployments.
+    ignore_changes = [
+      "site_config[0].linux_fx_version"
```
iphilpot: so this will set the fx_version on the first setup, then ignore this setting going forward?

infra/modules/providers/azure/app-service/variables.tf
```diff
@@ -58,7 +59,7 @@ variable "app_insights_instrumentation_key" {
 }
 
 variable "site_config_always_on" {
-  description = "Should the app be loaded at all times? Defaults to false."
+  description = "Should the app be loaded at all times? Defaults to true."
```
iphilpot: Why are we changing this?

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -6,6 +6,16 @@ locals {
...
   app_configs                    = values(var.app_service_config)
+
+  app_linux_fx_versions = [
+    for config in values(var.app_service_config) :
+    // Without specifyin a `linux_fx_version` the webapp created by the `azurerm_app_service` resource
+    // will be a non-container webapp.
+    //
+    // The value of "DOCKER" is a stand-in value that can be used to force the webapp created to be
+    // container compatible without explicitly specifying the image that the app should run.
+    config.image == "" ? "DOCKER" : format("DOCKER|%s/%s", var.docker_registry_server_url, config.image)
```
iphilpot: So if no image is provided we set this to docker? Then the test is expecting only "Docker" with no image? How does the end-to-end test with echo server change?

infra/modules/providers/azure/provider/main.tf
nmiodice: Can you elaborate on that @KeithJRome ?

infra/modules/providers/azure/app-service/main.tf
nmiodice: Yeah thats the long & short of it. If you specify an actual image to deploy it will configure that (on the first time only, since updates are ignored). If you don't, you get an "empty" app svc for containers.

infra/modules/providers/azure/app-service/variables.tf
nmiodice: We aren't, the docs were just wrong.

infra/modules/providers/azure/app-service/main.tf
nmiodice: No change in E2E. The `"DOCKER"` is simply to force the app service that is created to be a container service (as opposed to other non container types)

nmiodice: To elaborate, if you do specify an image (which is what the other templates do) then it will be deployed. hence no change to E2E tests.

infra/modules/providers/azure/provider/main.tf
iphilpot: @KeithJRome I read that thread as they are bumping the dependency to 0.12.8. Do you think that means anyone with the 1.34 provider needs to be running on 0.12.8? or is that only for the provider team? @nmiodice did you upgrade TF to 0.12.8? I didn't see that in the commits...

KeithJRome: I looked more closely into each of the changes since 1.32 and the one I was most interested in happens to be the same PR that fixes what you need. The others are very unlikely to cause us problems not that I've dug deeper into the actual code changes involved.

So I don't think this should cause us any problems. Carry On :)

KeithJRome: Yes @iphilpot, I think that means we need to rev our direct dependency on Terraform to 0.12.8, but I don't think this will cause us any problems. A few weird error messages might become less weird, and that's probably the extent of impact for us (aside from the bug that @nmiodice referred to being fixed).

nmiodice: @KeithJRome @iphilpot I don't think that we should be driving the version of Terraform based off of the version of the library that the providers consume as its not currently possible to satisfy each provider.

Null Resource Provider: https://github.com/terraform-providers/terraform-provider-null/blob/b8568b9b73a19322eaf0c44982f3834ff34854ff/go.mod#L5

Azure AD Resource Provider: https://github.com/terraform-providers/terraform-provider-azuread/blob/0ed698531d3dd13fed4e51469fda2d517f98aa9d/go.mod#L10

PR #335, opened 2019-09-27
Enable White-Listing of IP addresses and self-hosted build agents

ThePhen:
## All Submissions:

We're allowing IP address white listing resource attributes to be populated.

Issue Number: 242

- [YES]


devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -38,17 +38,17 @@ stages:
     configurationMatrix:
     - jobName: az_hello_world
       terraformTemplatePath: 'infra/templates/az-hello-world'
-      terraformWorkspacePrefix: 'hw'
+      terraformWorkspacePrefix: '242hw'
```
nmiodice:
```suggestion
      terraformWorkspacePrefix: 'hw'
```

devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -38,17 +38,17 @@ stages:
...
       environmentsToTeardownAfterRelease:
       - 'devint'
 
     - jobName: az_service_single_region
       terraformTemplatePath: 'infra/templates/az-service-single-region'
-      terraformWorkspacePrefix: 'sr'
+      terraformWorkspacePrefix: '242sr'
```
nmiodice:
```suggestion
      terraformWorkspacePrefix: 'sr'
```

devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -38,17 +38,17 @@ stages:
...
       environmentsToTeardownAfterRelease:
       - 'devint'
 
     - jobName: az_isolated_service_single_region
       terraformTemplatePath: 'infra/templates/az-isolated-service-single-region'
-      terraformWorkspacePrefix: 'iso'
+      terraformWorkspacePrefix: '242iso'
```
nmiodice:
```suggestion
      terraformWorkspacePrefix: '2so'
```
```suggestion
      terraformWorkspacePrefix: 'iso'
```

infra/templates/az-isolated-service-single-region/terraform.tfvars
```diff
@@ -30,7 +30,19 @@ unauthn_deployment_targets = [
 
 # Note: this is configured as such only to test IP Whitelists. This is a well
```
nmiodice: The comment was out-of-date before, but can we update it to reflect what these IPs are?

My *hunch* is that they are the build agent IPs that we are using in our build. However, its difficult to determine from the context alone.

infra/templates/az-isolated-service-single-region/tests/unit/unit_test.go
```diff
@@ -222,10 +157,8 @@ func TestTemplate(t *testing.T) {
 		TfOptions:             tfOptions,
 		Workspace:             workspace,
 		PlanAssertions:        nil,
-		ExpectedResourceCount: 62,
+		ExpectedResourceCount: 63,
 		ExpectedResourceAttributeValues: infratests.ResourceDescription{
-			// "module.keyvault.azurerm_key_vault.keyvault":                                   expectedKeyVault,
```
nmiodice: Are we explicitly removing both unit and integration tests that validate the IP whitelists?

Can you provide some context on why? I understand not having both, as it is redundant, but it would be nice to have at least one test that validates it is configured properly.

ThePhen: @nmiodice the IP addresses used as a part of the build are rather ephemeral, and I don't think they are relevant to int/unit testing.

The notion that we need to be able to test that setting an IP results in that IP being present in the white-list makes sense. We hit some problems with sorting out the check needed to work, however. 

Given the overall nature of the work, and the benefit to enabling the build agent pool, I'd like to treat the need to re-establish these particular tests as a new, small, targeted bug/issue.

ThePhen:
After talking with @nmiodice and @tasdevani21, we're going to close the PR, for the time being.

Issue #311, opened 2019-09-23
CICD Include PR number from build in the terraform workspace name to avoid deployment collisions during PR branch CI/CD processes


KeithJRome:
### Description
*As a developer, I'd like to have the CI/CD process perform any test deployments for my PR in a way that ensures no conflict with other PRs (or with master), in order to avoid any workspace or naming conflicts.*

It was determined that a good way to achieve this would be to modify our YAML pipeline to incorporate the PR number (if triggered via PR) into the workspace name. The workspace name is currently assembled by combining the `workspacePrefix` in the top-level YAML with the build environment (ie "devint"/etc).

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)
- [ ] CI/CD deployments from master branch should continue to operate the same as they do prior to this change (no change to workspace and resource naming)
- [ ] CI/CD deployments from Pull Request branches should insert the PR number into the workspace name (which then should cascade into the resource names)

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

ThePhen:
It might be important that someone/something clean-up AZ resources, once a PR is merged/closed (or else the resources will keep billing).

erikschlegel:
I love the idea of including the PR id within the workspace prefix.

PR #342, opened 2019-09-30
Append PR number to TF Workspace Name when available

KeithJRome:
## All Submissions:
* [NA] I have updated the documentation accordingly.

Individual PRs will compete for the same resource groups as the master branch deployment. This is a frequent immediate cause for many of the CD failures that we see in the PR checks for otherwise correct code. This problem is expected to become a more serious issue once a custom build agent is in use (with parallel builds).

Issue Number: #311

## What is the new behavior?
- When CI/CD checks are performed on a Pull Request, the PR number is appended to the Terraform Workspace Name. This ensures that the deployment of the PR that occurs during the checks will not race with other PR checks or with the main merge stream back to master.
- Master merges should continue to deploy as before, without an appended number in the workspace name.
- *As a [new] final step in the PR process, developers need to be mindful about tearing down these temporary workspaces once their code merges to master, in order to keep resource expenditure from spiraling.*

- [NO]

## Other information

### DIFF
--- a/devops/providers/azure-devops/templates/azure-pipelines.yml
+++ b/devops/providers/azure-devops/templates/azure-pipelines.yml
@@ -33,7 +33,9 @@ stages:
 - template: ./infrastructure/azure-pipeline-cicd-compose.yml
   parameters:
     environments:
-    - 'devint'
+    - name: 'devint'
+      enablePrIsolation: true
+      resourceNameIsolationLevel: 8
 
     configurationMatrix:
     - jobName: az_hello_world
--- a/devops/providers/azure-devops/templates/infrastructure/scripts/compute-workspace.yml
+++ b/devops/providers/azure-devops/templates/infrastructure/scripts/compute-workspace.yml
@@ -1,6 +1,7 @@
 parameters:
   terraformWorkspacePrefix: ''
   environmentName: ''
+  enablePrIsolation: true
   stepName: ''
 
 steps:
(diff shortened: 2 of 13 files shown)


devops/providers/azure-devops/templates/azure-pipelines.yml
```diff
@@ -33,7 +33,9 @@ stages:
 - template: ./infrastructure/azure-pipeline-cicd-compose.yml
   parameters:
     environments:
-    - 'devint'
+    - name: 'devint'
+      enablePrIsolation: true
```
iphilpot: Should this come from a variable in the var group?

devops/providers/azure-devops/templates/infrastructure/scripts/compute-workspace.yml
```diff
@@ -0,0 +1,58 @@
...
+        # produces a simple combination of prefix+environment to be used as the workspace name.
+
+        PR_NUMBER=""
+        if [[ $PR_ISOLATION == true ]]; then
+
+            # BUILD_SOURCEBRANCH will look like "refs/pull/1/merge" if this is a PR branch
+            IFS='/' # delimiter
+            read -ra PARTS <<< "$BUILD_SOURCEBRANCH" # split the branch into parts
+
+            if [[ "${#PARTS[@]}" == "4" && "${PARTS[0]}" == "refs" && "${PARTS[1]}" == "pull" && "${PARTS[3]}" == "merge" ]]; then
```
iphilpot: Nice check... sed/awk could be used here, but this does the trick.

devops/providers/azure-devops/templates/azure-pipelines.yml
KeithJRome: Good idea. If it's useful to do that, then we should spin off a post-v1 feature request story for it.

Issue #315, opened 2019-09-24
Authenticed App Service only applies auth policy to the production slot (not staging)


KeithJRome:
### Description
*As a developer, I'd expect the staging slot that is created for an authenticated app service to be configured for auth (not just the production slot), in order to allow validation of newly deployed containers prior to rotation into production.*

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)
- [x] Authenticated app service should have authentication enabled for the default Production slot
- [x] Authenticated app service should have authentication enabled for the Staging slot

### Resources
- Documentation for the `--slot` parameter indicates that the default is to apply to the Production slot only: https://docs.microsoft.com/en-us/cli/azure/webapp/auth?view=azure-cli-latest#az-webapp-auth-update
- CLI code invoked in provisioner for app service does not include a second call to `az webapp auth update`, which would be needed to configure the staging slot as well: `cobalt/infra/templates/az-isolated-service-single-region/ase.tf`

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

PR #344, opened 2019-10-01
Add easy auth to app service slots

TechnicallyWilliams:
## All Submissions:
* [NO] All new and existing tests passed.

Issue Number: #315 "Authenticated app service only applies auth settings to the production slot (not staging)"

## What is the new behavior?

Authenticated app service now applies authentication to the staging slots of any Authenticated app service.

- [NO]

**Configuration plus output:**
image

## Other information

Configuration results in a single authentication service for each app and it's related app slots.

### DIFF
--- a/infra/modules/providers/azure/ad-application/main.tf
+++ b/infra/modules/providers/azure/ad-application/main.tf
@@ -18,6 +18,7 @@ resource "azuread_application" "auth" {
 
 # Gives us access to outputs not directly provided by the resource
 data "azuread_application" "auth" {
-  count     = length(var.ad_app_config)
-  object_id = azuread_application.auth[count.index].object_id
+  count       = length(var.ad_app_config)
+  depends_on  = [azuread_application.auth]
+  object_id   = azuread_application.auth[count.index].object_id
 }
--- a/infra/modules/providers/azure/ad-application/main.tf
+++ b/infra/modules/providers/azure/ad-application/main.tf
@@ -18,7 +18,7 @@ resource "azuread_application" "auth" {
 
 # Gives us access to outputs not directly provided by the resource
 data "azuread_application" "auth" {
-  count       = length(var.ad_app_config)
-  depends_on  = [azuread_application.auth]
-  object_id   = azuread_application.auth[count.index].object_id
+  count      = length(var.ad_app_config)
+  depends_on = [azuread_application.auth]
+  object_id  = azuread_application.auth[count.index].object_id
 }
--- a/infra/modules/providers/azure/app-service/main.tf
+++ b/infra/modules/providers/azure/app-service/main.tf
@@ -106,6 +106,7 @@ data "azurerm_app_service" "all" {
   count               = length(azurerm_app_service.appsvc)
   name                = azurerm_app_service.appsvc[count.index].name
   resource_group_name = data.azurerm_resource_group.appsvc.name
+  depends_on          = [azurerm_app_service.appsvc]
 }
 
 resource "azurerm_template_deployment" "access_restriction" {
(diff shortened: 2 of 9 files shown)


infra/modules/providers/azure/ad-application/variables.tf
```diff
@@ -7,6 +7,11 @@ variable "ad_app_config" {
   }))
 }
 
+variable "instance_count" {
```
iphilpot: Why are we making this a configuration element? I think this could add confusion...

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -103,9 +103,10 @@ resource "azurerm_app_service_slot" "appsvc_staging_slot" {
 }
 
 data "azurerm_app_service" "all" {
-  count               = length(azurerm_app_service.appsvc)
+  count               = length(azurerm_app_service_slot.appsvc_staging_slot)
```
iphilpot: how do we know how many app services there are now?

infra/modules/providers/azure/ad-application/variables.tf
TechnicallyWilliams: The feature in this PR prompted me to make an app service output configuration change which required a map that mapped app services to their respective slots. The ad-application relies on this new mapping, however, the map is so dynamic that terraform couldn't infer a count for the ad-application to rely on at plan time ("compile time") and would throw a error. My response to that error  prompted to bring back ad-application's "instance count" variable.

infra/modules/providers/azure/app-service/main.tf
TechnicallyWilliams: Glad you saw this! You're right, this data block drives the last and final output of the iso template. I'll revert this back and look for any of the consequences to that.

iphilpot: @TechnicallyWilliams any update on this?

TechnicallyWilliams: Yeah, i'm making changes locally at the moment. May need to rebase with master as well.

infra/modules/providers/azure/ad-application/variables.tf
TechnicallyWilliams: Found a simple way to keep "instance count" out of the picture. Thanks for calling this out. Will update shortly!

TechnicallyWilliams: Instance count has been removed. Also, screenshot of staging confi and auth login screen is now included in PR description.

infra/modules/providers/azure/ad-application/main.tf
```diff
@@ -18,6 +18,7 @@ resource "azuread_application" "auth" {
 
 # Gives us access to outputs not directly provided by the resource
 data "azuread_application" "auth" {
-  count     = length(var.ad_app_config)
-  object_id = azuread_application.auth[count.index].object_id
+  count      = length(var.ad_app_config)
+  depends_on = [azuread_application.auth]
```
nmiodice: I think this is unnecessary based on the implicit dependency that is created on line 23

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -106,6 +106,7 @@ data "azurerm_app_service" "all" {
   count               = length(azurerm_app_service.appsvc)
   name                = azurerm_app_service.appsvc[count.index].name
   resource_group_name = data.azurerm_resource_group.appsvc.name
+  depends_on          = [azurerm_app_service.appsvc]
```
nmiodice: Similar comment r.e. the implicit dependency created on line 107

Issue #330, opened 2019-09-26
Ignore Key Vault Secrets after initial create


iphilpot:
### Description
As an engineer, I'd like to ensure that the secrets in key vault can change from outside of TF and not have TF recreate them, in order to allow for key rotation.

We should add the ignore lifecycle [or something similar] to these items and possibly the AD SP/App Registration items. We probably will need to look deeper into this and how the two interact.

### Acceptance Criteria
Reference: [Done-Done Checklist] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/DoneDone.md)
- [ ] Should only ever create the secrets once
- [ ] Should not affect the AD App Registrations lifecycle
- [ ] Should not affect the AD Service Principle lifecycle

Also, here are a few points that need to be addressed:

1. Constraint 1;
1. Constraint 2;
1. Constraint 3.

### Resources
*Technical Design Document*
*Mockups*

### Tasks
*Stories are intended to be completed in a single sprint; if task breakdown creates addition work then team should discuss promoting the Story to an Epic.
Reference: [Minimal Valuable Slices] (https://github.com/Microsoft/code-with-engineering-playbook/blob/master/Engineering/BestPractices/MinimalSlices.md)*

Reference: [How to Write Better Tasks] (http://agilebutpragmatic.blogspot.com/2012/04/splitting-story-into-tasks-how-to-write.html)

**_Assignee should break down work into tasks here_**

PR #351, opened 2019-10-02
Ignore service principal password resources from Terraform state management

erikschlegel:
## All Submissions:
* [YES/NO/NA] My code follows the code style of this project.

Currently, all azure resource(s) state changes related to the service principal password are managed through Terraform. This results in state changes being triggered for Terraform incremental builds as the password is randomly generated on each run. The impact is Terraform attempts to delete and recreate the password and KV secrets on each incremental deployment, regardless of template changes. Terraform should only be responsible for creating resources like the SP password and keyvault secrets when bootstrapping the initial azure environment. 

Issue Number: #330 

## What is the new behavior?
Terraform resource types such as service prinipal passwords and keyvault secrets capturing sp passwords are ignored from state change lifecycles events via 
```
lifecycle {
    ignore_changes = ["value"]
  }
```

- [NO]


erikschlegel:
@iphilpot all checks are passing and ready for review

infra/modules/providers/azure/service-principal/main.tf
```diff
@@ -28,4 +28,8 @@ resource "azuread_service_principal_password" "sp" {
   service_principal_id = azuread_service_principal.sp[0].object_id
   value                = local.sp_password
   end_date_relative    = var.sp_pwd_end_date_relative
+
+  lifecycle {
```
TechnicallyWilliams: These lifecycle hooks bake in the original passwords inserted into the creation of the service principal password and the key vault secrets that store them. How will password updates be managed due to this change? Outside of terraform?

iphilpot: Yes should be completed outside of TF

iphilpot: Any reason to put this in the SP module and not in the template?

iphilpot: nevermind, it's a config element it looks like.

PR #396, opened 2020-07-16
feat: Upgrade azurerm provider to 2.9

helayoty:
## All Submissions:

The current azurerm provider version (1.4) will not work for new features introduced by > 2.0.0 and to be able to contribute back specific modules such as azure ML workspace, we needed to pump the azurerm provider version to >= 2.6

We had to go with version 2.9 as it has fix for app gateway issues.

Issue Number: N/A

## What is the new behavior?

- Update provider module 
- Update backend files for templates
- Update keyvault module
- Update app-service module

- [YES]

- Had to update keyvault module by using "data.azurerm_client_config.current.object_id" instead of "data.azurerm_client_config.current.service_principal_object_id"
- Had to remove "virtual_network_name" from app-service module 

## Other information

### DIFF
--- a/infra/modules/providers/azure/app-service/main.tf
+++ b/infra/modules/providers/azure/app-service/main.tf
@@ -58,10 +58,9 @@ resource "azurerm_app_service" "appsvc" {
   )
 
   site_config {
-    linux_fx_version     = local.app_linux_fx_versions[local.app_names[count.index]]
-    always_on            = var.site_config_always_on
-    virtual_network_name = var.vnet_name
-    app_command_line     = var.app_service_config[local.app_names[count.index]].app_command_line
+    linux_fx_version = local.app_linux_fx_versions[local.app_names[count.index]]
+    always_on        = var.site_config_always_on
+    app_command_line = var.app_service_config[local.app_names[count.index]].app_command_line
   }
 
   identity {
--- a/infra/modules/providers/azure/app-service/main.tf
+++ b/infra/modules/providers/azure/app-service/main.tf
@@ -95,10 +95,9 @@ resource "azurerm_app_service_slot" "appsvc_staging_slot" {
   )
 
   site_config {
-    linux_fx_version     = local.app_linux_fx_versions[local.app_names[count.index]]
-    always_on            = var.site_config_always_on
-    virtual_network_name = var.vnet_name
-    app_command_line     = var.app_service_config[local.app_names[count.index]].app_command_line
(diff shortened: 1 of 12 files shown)


helayoty:
@nmiodice

infra/modules/providers/azure/provider/main.tf
```diff
@@ -1,5 +1,6 @@
 provider "azurerm" {
-  version = "~>1.40.0"
+  version = "~> 2.6.0"
```
nmiodice, line 2: woot!

infra/templates/az-hello-world/backend.tf
```diff
@@ -4,3 +4,8 @@ terraform {
   }
 }
 
+provider "azurerm" {
```
nmiodice, line 7: Do we need to change these? I thought we pulled the provider in from `/infra/modules/providers/azure/provider`?

infra/modules/providers/azure/app-service/main.tf
```diff
@@ -58,10 +58,9 @@ resource "azurerm_app_service" "appsvc" {
   )
 
   site_config {
-    linux_fx_version     = local.app_linux_fx_versions[local.app_names[count.index]]
-    always_on            = var.site_config_always_on
-    virtual_network_name = var.vnet_name
```
nmiodice, line 63: This is actually a breaking change. It does look like its not supported anymore in the new TF provider, but I wonder if there is a different way that we can configure it?

The impact is that these templates are no longer VNET isolated.

infra/templates/az-hello-world/backend.tf
helayoty: for a reason, I didn't discover it yet, we have to add the provider in each backend.tf with the features {} expression. Otherwise, we'll get an error

infra/modules/providers/azure/app-service/main.tf
helayoty: We sill have a VNET isolated app-service by using **resource "azurerm_template_deployment"** we are using [here](https://github.com/microsoft/cobalt/blob/master/infra/modules/providers/azure/app-service/main.tf#L135)

infra/templates/az-hello-world/backend.tf
nmiodice: Ugh, ok. No need to change, but that is too bad!

helayoty: it is indeed!

